PR #1, opened 2019-11-18
Added a way to expose a public key by validating a signed message.

howech:
Currently only works for p2pkh addresses. 

This makes it possible to use caravan to create multisig addresses that include OpenDime keys as part of the signature. (Spending protocol still TBD.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1631,6 +1631,19 @@
         "wif": "^2.0.1"
       }
     },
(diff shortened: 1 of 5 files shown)


waldenraines:
@dhruvbansal @howech is this still change still desired?

waldenraines:
Closing due to inactivity.

PR #2, opened 2019-11-23
Option to not validate output amount greater than input amount

rsbondi:
This is a valid state for a wallet doing coin selection, when a user adds outputs and before the wallet does coin selection, the output amount will be more than the input amount until enough coins are selected.

By making `inputsTotalSats` and optional parameter to `validateOutputAmountBTC` the undesired message is ignored if `inputsTotalSats` is omitted.

### DIFF
--- a/src/transactions.js
+++ b/src/transactions.js
@@ -1,4 +1,4 @@
-/** 
+/**
  * This module provides validation messages related to transactions.
  * @module transactions
  */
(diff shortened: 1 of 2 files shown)


rsbondi:
This is now handled [here](https://github.com/unchained-capital/unchained-bitcoin/blob/e9177e7171c3bf11ad5c245c7d633dac484f409c/src/outputs.js#L129) so closing

PR #5, opened 2020-02-29
Convert extended public keys between formats

rsbondi:
This will convert from any format, xpub, ypub, zpub etc to any other format.  Additional validation is added as a check to see if a conversion is needed.


src/keys.js
```diff
@@ -11,13 +11,88 @@ import {validateHex, toHexString} from "./utils";
 import {TESTNET, networkData} from "./networks";
 
 const bip32 = require('bip32');
+const bs58check = require('bs58check');
+
+export const extendedPublicKeyVersions = {
```
dhruvbansal, line 16: All caps might be better for this: `EXTENDED_PUBLIC_KEY_VERSIONS`

waldenraines:
@dhruvbansal is this still change still desired?

waldenraines:
Closing due to inactivity.

PR #6, opened 2020-03-05
docs: add build scripts for publishing docs on gh-pages

bucko13:
implements same changes as implemented in unchained-wallets to remove docs from `master` git history. 

Will need change in github settings to publish pages from` gh-pages` rather than master /docs folder


howech:
Please move the script to a bin directory and update the .npmignore file.

PR #10, opened 2020-04-15
add ExtendedPublicKey Class for easy encoding and decoding

bucko13:
the bip32 package dependency ended up introducing a breaking change via some upstream package update. It was becoming complex to shoehorn xpub encoding using that package so I made a local class in the keys.js file called `ExtendedPublicKey` that does some type checking on arguments when wanting to encode an xpub, and supports decoding it as well. 

Some other linting updates got picked up in the process.


package.json
```diff
@@ -40,8 +40,11 @@
   "dependencies": {
     "@babel/polyfill": "^7.7.0",
     "bignumber.js": "^8.1.1",
+    "bip32": "^2.0.5",
```
bucko13, line 43: This was being used but via another dependency. I wanted to be explicit about our usage to be sure of the version and API we were using.

package.json
```diff
@@ -40,8 +40,11 @@
...
     "bitcoin-address-validation": "^0.2.9",
     "bitcoinjs-lib": "^4.0.5",
+    "bs58check": "^2.1.2",
```
bucko13, line 46: This was also an implicit dependency that is now explicit.

package.json
```diff
@@ -40,8 +40,11 @@
...
+    "bufio": "^1.0.6",
```
bucko13, line 47: This is for using a JS `Struct` object as well as Buffer reader/writer. (from the bcoin-org project)

src/keys.js
```diff
@@ -33,10 +36,117 @@ export const EXTENDED_PUBLIC_KEY_VERSIONS = {
...
+ * @param {string} options.parentFingerprint - fingerprint of parent public key
+ * @param {string} [options.network = mainnet] - mainnet or testnet
+ * @example
+ * import { ExtendedPublicKey } from "unchained-bitcoin"
+ * const xpub = ExtendedPublicKey.fromBase58("xpub6CCHViYn5VzKSmKD9cK9LBDPz9wBLV7owXJcNDioETNvhqhVtj3ABnVUERN9aV1RGTX9YpyPHnC4Ekzjnr7TZthsJRBiXA4QCeXNHEwxLab")
+ * console.log(xpub.encode()) // returns raw Buffer of xpub encoded as per BIP32
+ * console.log(xpub.toBase58()) // returns base58 check encoded xpub
+ */
+export class ExtendedPublicKey extends Struct {
+  constructor(options) {
```
bucko13, line 61: We do a lot of type checks in the constructor to make sure we're getting the arguments in the right format (some are expected to be ints, some strings. Some strings we can also know what length we expect them to be.

src/keys.js
```diff
@@ -33,10 +36,143 @@ export const EXTENDED_PUBLIC_KEY_VERSIONS = {
...
+ * @param {string} options.chaincode chaincode corresponding to pubkey and path
+ * @param {string} options.parentFingerprint - fingerprint of parent public key
+ * @param {string} [options.network = mainnet] - mainnet or testnet
+ * @example
+ * import { ExtendedPublicKey } from "unchained-bitcoin"
+ * const xpub = ExtendedPublicKey.fromBase58("xpub6CCHViYn5VzKSmKD9cK9LBDPz9wBLV7owXJcNDioETNvhqhVtj3ABnVUERN9aV1RGTX9YpyPHnC4Ekzjnr7TZthsJRBiXA4QCeXNHEwxLab")
+ * console.log(xpub.encode()) // returns raw Buffer of xpub encoded as per BIP32
+ * console.log(xpub.toBase58()) // returns base58 check encoded xpub
+ */
+export class ExtendedPublicKey extends Struct {
```
bucko13, line 60: Using the `Struct` interface gives us a lot of built in tools including encoding and decoding methods and a `clone` method. We're building on top of this to have a `toBase58` and `fromBase58` interface to easily retrieve the xpub given a set of data about a BIP32 node. This could be expanded upon in the future to have a more robust BIP32 interface and break dependence on the bitcoinjs implementation.

src/keys.test.js
```diff
@@ -1,4 +1,5 @@
-import bs58 from 'bs58'
+import bs58check from "bs58check";
```
bucko13, line 1: A lot of linting changes in this file but the bottom has new tests.

PR #11, opened 2020-05-01
include marker and version bytes in fee estimation for P2WSH

bucko13:
This gets to the correct calculation for a 2-of-3 multisig with 2 inputs and 2 outputs as calculated by the bitcoind rpc and estimated at blockstream.info. The new test I added was failing before this change. 

I wasn't too sure how the `overhead` value is derived, but it did look like based on the bitcoin core docs [here](https://bitcoincore.org/en/segwit_wallet_dev#transaction-fee-estimation) that new segwit fields for marker and flag needed to be included in the witness part of the calculation (i.e. in the discount). Adding that seemed to fix the calcluation.

### DIFF
--- a/src/p2wsh.js
+++ b/src/p2wsh.js
@@ -1,3 +1,5 @@
+import {encoding} from 'bufio';
+
 /**
  * This module provides functions and constants for the P2WSH address type.
  * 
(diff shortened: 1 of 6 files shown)


bucko13:
I've added a lot more detail to the calculation of the tx size estimate and modularized for each part of the tx itself. This was helpful for me to make sure each piece was being calculated correctly, but also should help future developers understand the calculation better. 

Several of these pieces could probably be abstracted out into a reusable module with an option passed in for deciding which type of tx we're calculating. But in the interest of not poking too many different pieces, I left all in `pswsh` for now. Happy to further refactor though.

src/p2wsh.js
```diff
@@ -14,11 +16,140 @@
...
+ * vsize is equal three times "base size" of a tx w/o witness data, plus the
+ * total size of all data, with the final result divided by scaling factor
+ * of 4 and round up to the next integer. For example, if a transaction is
+ * 200 bytes with new serialization, and becomes 99 bytes with marker, flag,
+ * and witness removed, the vsize is (99 * 3 + 200) / 4 = 125 with round up.
+ * @param {Number} baseSize - base size of transaction
+ * @param {Number} witnessSize - size of witness fields
+ * @returns {Number} virtual size of tx
+ */
+function calculateVSize(baseSize, witnessSize) {
```
bucko13, line 143: This could be reused for other esitmators.

src/p2wsh.js
```diff
@@ -14,11 +16,140 @@
...
+
+/**
+ * @description Calculates the size of the fields in a transaction which DO NOT
+ * get counted towards witness discount.
+ * Calculated as: version bytes (4) + locktime bytes (4) + input_len (1+) + txins (41+) + output_len (1+) + outputs (9+)
+ * @param {*} inputsCount - number of inputs in the tx
+ * @param {*} outputsCount - number of outputs in the tx
+ * @returns {Number} number of bytes in the tx without witness fields
+ */
+export function calculateBase(inputsCount, outputsCount) {
```
bucko13, line 108: If used in other contexts (namely P2SH-P2WSH/P2WPKH) this would need to be paramtarized for larger txins, but other fields are the same.

src/p2wsh.js
```diff
@@ -14,11 +16,140 @@
...
+/**
+ * @description Calculates the value of a multisig witness given m-of-n values
+ * Calculation is of the following form:
+ * witness_items count (varint 1+) + null_data (1 byte) + size of each signature (1 byte * OP_M) + signatures (73 * M) +
+ * redeem script length (1 byte) + redeem script size (4 + 34 * N bytes)
+ * @param {Number} m - value of m in m-of-n for multisig script
+ * @param {Number} n - value of n in m-of-n for multisig script
+ * @returns {Number} 6 + (74 * M) + (34 * N)
+ */
+export function getWitnessSize(m, n) {
```
bucko13, line 82: This stays the same for both types of P2WSH, but would need to be adjusted for P2WPKH support.

src/p2wsh.js
```diff
@@ -14,11 +16,140 @@
...
+/**
+ * @description calculates size of redeem script given n pubkeys.
+ * Calculation looks like: 
+ * OP_M (1 byte) + size of each pubkey in redeem script (OP_DATA= 1 byte * N) +
+ * pubkey size (33 bytes * N) + OP_N (1 byte) + OP_CHECKMULTISIG (1 byte)
+ *  => 1 + (1 * N) + (33 * N) + 1 + 1
+ * @param {Number} n - value of n in m-of-n for multisig script
+ * @returns {Number} 3 + 34 * N
+ */
+export function getRedeemScriptSize(n) {
```
bucko13, line 64: This could be reused for other P2SH/P2WSH estimators.

src/p2wsh.js
```diff
@@ -14,11 +16,140 @@
  */
 export const P2WSH = "P2WSH";
 
+/**
+ * @description provides the size of single tx input for a segwit tx (i.e. empty script)
+ * Each input field will look like:
+ * prevhash (32 bytes) + prevIndex (4) + scriptsig (1) + sequence bytes (4)
+ * @returns {Number} 41 (always 41 for segwit inputs since script sig is in witness)
+ */
+function txinSize() {
```
bucko13, line 25: This could be parameterized to support P2SH-P2WSH transactions.

src/p2wsh.js
```diff
@@ -14,11 +16,140 @@
...
+ */
+export function calculateBase(inputsCount, outputsCount) {
+  let total = 0;
+  total += 4; // version
+  total += 4 // locktime
+
+  total += encoding.sizeVarint(inputsCount); // inputs length
+  total += inputsCount * txinSize();
+  total += encoding.sizeVarint(outputsCount); 
+  total += outputsCount * txoutSize();
```
destrys, line 116: Is this assuming that all outputs are p2wsh?

src/p2wsh.test.js
```diff
@@ -1,15 +1,47 @@
...
         m: 2,
         n: 3})).toBe(202); // actual value from bitcoin core for P2PKH out
     });
+    const vsize = estimateMultisigP2WSHTransactionVSize({
+      numInputs: 2,
+      numOutputs: 2,
+      m: 2,
+      n: 3
+    })
+    // from actual p2wsh payment with vsize 306
```
destrys, line 23: Include the txid in the comment.

src/p2wsh.js
bucko13: Yes, exactly. I mention this in the txoutSize function but this is something else that could probably be paramterized: https://github.com/unchained-capital/unchained-bitcoin/pull/11/files/2d67a90045f60ba82a46ae8bc105860869714734#diff-bb07f41b65bd25e15d271ea48baaefe1R42

PR #17, opened 2020-08-31
Adds `fingerprintToFixedLengthHex` function and tests

humanumbrella:
Given the existing fingerprint functionality, here we add a function to output a fixed-length, zero-padded, hex string that is exactly eight characters in length.

Unit tests included. Updated some of the jsdocs as well.

### DIFF
--- a/src/keys.test.js
+++ b/src/keys.test.js
@@ -14,6 +14,7 @@ import {
   validatePrefix,
   EXTENDED_PUBLIC_KEY_VERSIONS,
   ExtendedPublicKey,
+  fingerprintToFixedLengthHex,
 } from "./keys";
 
 import { TESTNET, MAINNET } from "./networks";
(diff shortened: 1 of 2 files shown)


src/keys.test.js
```diff
@@ -384,6 +385,30 @@ describe("keys", () => {
...
+        '2371': '00000943',
+        '53947': '0000d2bb',
+        '322639': '0004ec4f',
+        '16033212': '00f4a5bc',
+        '25201234': '01808a52',
+        '849849884': '32a7ae1c',
+        '8498490984': 'fa8cae68', // should be 1fa8cae68 but truncated front 1
+      };
+
+      for (const xfp of Object.keys(hexOutputs)) {
```
destrys, line 402: Instead of keys, you can key both the key and value using`Object.entries`.
And if you flip the key/values then you could avoid the `Number` coersion.
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

destrys: e.g.:
```      
const hexOutputs = {
        '00000007': 7,
...
```

destrys (dismissed):
Left a comment for improvement, but LGTM.

src/keys.test.js
humanumbrella: Good call

waldenraines:
@destrys anything else you'd like to see?

destrys (approved):
Nice!

destrys:
LGTM, go for it.

PR #20, opened 2020-09-22
Remove dead code

dergigi:
The `bitcoinAddressValidation` call below does a validity check for all address types. E.g. see https://github.com/ruigomeseu/bitcoin-address-validation/blob/master/src/index.js#L36 and https://github.com/bitcoinjs/bech32/blob/master/index.js#L106 for bech32.

### DIFF
--- a/src/addresses.js
+++ b/src/addresses.js
@@ -46,13 +46,6 @@ export function validateAddress(address, network) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


waldenraines (approved):
LGTM, thanks @dergigi!

PR #22, opened 2020-11-26
PSBT creation and Braid support

humanumbrella:
- create a concept of a 'Braid' which is a set of ExtendedPublicKeys
- Create methods for creating PSBTs based on inputs that are aware of their braid (e.g. which extended public keys are they associated with?)
- PSBT typedefs into psbt.js
- adds validateMultisigInputs(input, braidRequired=true) option to validate inputs have knowledge of their braid.
- updates all fixtures to include transactions that have change (accomplish this by setting fee to 1sat/byte and going through with the existing transaction.)
- adds psbts (unsigned and partially signed) for each multisig transaction type
- small edits to other tests to get code coverage to 100%
- bumps walletConfigs in fixtures to show v0.0.2 because Coldcard firmware will start requiring the Derivation property.


dhruvbansal (changes requested):

src/psbt.js
```diff
@@ -8,10 +22,230 @@ import {toHexString} from './utils';
...
+ * ADDITIONAL NOTE: because of the restrictions we place on braids to march their
+ * multisig addresses (slices) forward at the *same* index across each chain of the
+ * braid, we do not run into a possible collision with this data structure.
+ * BUT - to have this method accommodate the *most* general form of signature parsing,
+ * it would be wise to wrap this one level deeper like:
+ *
+ *                     address: [pubkey : [signature(s)]]
+ *
+ * that way if your braid only advanced one chain's (member's) index so that a pubkey
+ * was could be used in more than one address, everything would still function properly.
```
bucko13, line 247: typo? Delete `was` I think.

src/transactions.js
```diff
@@ -81,6 +85,48 @@ export function unsignedMultisigTransaction(network, inputs, outputs) {
...
+ * and outputs stored as a PSBT object
+ *
+ * Returns a [`PSBT`]{@link https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/types/psbt.d.ts|PSBT} object from bitcoinjs-lib.
+ *
+ * @param {module:networks.NETWORKS} network - bitcoin network
+ * @param {module:inputs.MultisigTransactionInput[]} inputs - transaction inputs : NOTE - must be braid-aware
+ * @param {module:outputs.TransactionOutput[]} outputs - transaction outputs
+ * @returns {Psbt} an unsigned bitcoinjs-lib Psbt object
+ */
+export function unsignedMultisigPSBT(network, inputs, outputs) {
```
bucko13, line 99: I don't know the answer to this, so just asking, but would it make sense to just have this accept a multisig object? Don't all of these exist on the multisig or am I wrong to think that? Almost could be a method on a multisig object itself. `multisig.getUnsignedPSBT`

src/transactions.js
```diff
@@ -191,6 +234,35 @@ export function signedMultisigTransaction(network, inputs, outputs, transactionS
...
+//  * Create a fully signed multisig transaction based on the unsigned
+//  * transaction, braid-aware inputs, and their signatures.
+//  *
+//  * @param {module:networks.NETWORKS} network - bitcoin network
+//  * @param {module:inputs.MultisigTransactionInput[]} inputs - braid-aware multisig transaction inputs
+//  * @param {module:outputs.TransactionOutput[]} outputs - transaction outputs
+//  * @param {Object[]} transactionSignatures - array of transaction signatures, each an array of input signatures (1 per input)
+//  * @returns {Transaction} a signed {@link https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/types/transaction.d.ts} Transaction object
+//  */
+ // export function signedMultisigPSBT(network, inputs, outputs, transactionSignatures) {
```
bucko13, line 247: Is this just a todo for later?

bucko13 (changes requested):

bucko13:
@dhruvbansal pinging you for feedback or comments on the review I left.

src/addresses.js
```diff
@@ -38,6 +38,7 @@ export function validateAddress(address, network) {
   const isBech32 = address.match(BECH32_ADDRESS_MAGIC_BYTE_REGEX);
   const address_body_regex = (isBech32 ? BECH32_ADDRESS_BODY_PATTERN : ADDRESS_BODY_PATTERN);
   const address_regex = magic_byte_regex + address_body_regex;
+  // This tests whether you've got the network lined up with address type or not
   if (! address.match(address_regex)) {
     if (network === TESTNET) {
       return "Address must start with one of 'tb1', 'm', 'n', or '2' followed by letters or digits.";
```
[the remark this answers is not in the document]
humanumbrella: Yes, but not here. Need to *convert* an address before asking the node about it

src/fixtures.js
```diff
@@ -285,21 +699,71 @@ const MULTISIGS_BASE = [
           "02000000000101eef67cc41c10722be710952296866ea13ed1608acdf15e453e8d874e6a15c6d50000000017160014a989c1c6a3dbbf44d508d5f36df2d08c97e9fca4feffffff02a08601000000000017a914dac0270cbf87a65c0cf4fd2295eb44c756b288ec87a5b90d000000000017a914d0324b98895786d859ae3ee3df0c384249f1a4ab870247304402200621a08b242b807a0c39b6e0bf302e503b9a2596acdd218b176cb62afba31824022027bb2cc91b5ae57500a2e3c440d4d45cc42617f4db665a1b25974122b3789ddd01210209bb437e2e4658c6eb92c20a1ef459a2d1da50757dfba0c49a19dd3dbd621d87e09b1800",
       },
     ],
+    // SINGLE OUTPUT
```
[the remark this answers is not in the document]
humanumbrella: ya, I originally left all the old transactions (no change outputs) in there ... but I can remove - it'll be in git history

src/keys.js
```diff
@@ -458,6 +502,28 @@ export function fingerprintToFixedLengthHex(xfp) {
   return (xfp + 0x100000000).toString(16).substr(-8);
 }
 
+/**
+ * Returns the root fingerprint of the extendedPublicKey
+ * (It should be zero-padded, hex-formatted string that is exactly eight
+ * characters long.)
```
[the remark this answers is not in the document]
humanumbrella: We will rely on the validation when the struct is created if it gets added.

src/psbt.js
```diff
@@ -8,10 +22,230 @@ import {toHexString} from './utils';
...
+ *
+ * const braid = generateBraidFromExtendedPublicKeys(MAINNET, P2SH, {{'xpub...', bip32path: "m/45'/0'/0'"}, {'xpub...', bip32path: "m/45'/0/0"}, {'xpub...', bip32path: "m/45'/0/0"}}, 2);
+ * const multisig = braid.deriveMultisigByIndex("0");
+ * console.log(getBip32Derivation(multisig, 0)); // {
+ *   {masterFingerprint: Buffer('1234..', 'hex'), path: "m/45'/0'/0'/0/0", pubkey: Buffer.from("02...", 'hex')}
+ *   {masterFingerprint: Buffer('3453..', 'hex'), path: "m/45'/0/0/0/0", pubkey: Buffer.from("03...", 'hex')}
+ *   {masterFingerprint: Buffer('1533..', 'hex'), path: "m/45'/0/0/0/0", pubkey: Buffer.from("02...", 'hex')}
+ * }
+ */
+function getBip32Derivation(multisig, index= 0) {
```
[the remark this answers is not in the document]
humanumbrella: It's not used anywhere outside of this file though. `Bip32Derivation` is a psbt-specific concept. Unfortunately, the name isn't as nice when it's multisig with many (potentially completely unrelated) bip32derivations for each multisig ... heh.

src/transactions.js
humanumbrella: No - it would have to support a potential array of multisigs, and the outputs are not always multisigs (only for change). So I'd say that the method signature should stay as is. Maybe if you wanted to generate a PSBT for a single UTXO in a slice, perhaps you might want to do that? Not sure

src/transactions.js
humanumbrella: Yes, will add TODO

dhruvbansal (approved):

PR #25, opened 2021-01-12
extend validatePublicKey to accept addressType parameter

bruce1971:
In this small PR, we extend the functionality of the validatePublicKey function, by allowing it to accept an addressType parameter.

As we know, uncompressed public keys in general are valid public keys...
However uncompressed public keys are not valid for P2WSH and P2SH-P2WSH! 

This extension of the validatePublicKey function will be used in [THIS PR](https://github.com/unchained-capital/caravan/pull/205) to fix [THIS ISSUE](https://github.com/unchained-capital/caravan/issues/164)

NOTE: validatePublicKey will of course be backwards compatible. ie if no addressType is supplied, validatePublicKey will function as previously! The test cases confirm this.

### DIFF
--- a/src/keys.js
+++ b/src/keys.js
@@ -332,7 +332,7 @@ export function validateExtendedPublicKey(xpubString, network) {
  * - Must be a valid BIP32 public key.
  *
  * @param {string} pubkeyHex - (compressed) public key in hex
- * @param {string} addressType - one of P2SH, P2SH-P2WSH, P2WSH
+ * @param {string} addressType - (optional) one of P2SH, P2SH-P2WSH, P2WSH
  * @returns {string} empty if valid or corresponding validation message if not
  * @example
  * import {validatePublicKey} from "unchained-bitcoin";
@@ -342,7 +342,7 @@ export function validateExtendedPublicKey(xpubString, network) {
  * console.log(validatePublicKey("03b32dc780fba98db25b4b72cf2b69da228f5e10ca6aa8f46eabe7f9fe22c994ee")); // ""
  * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53")); // ""
  * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53", "P2SH")); // ""
- * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53", "P2WSH")); // "P2WSH doesnt support uncompressed public key."
+ * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53", "P2WSH")); // "P2WSH does not support uncompressed public key."
  */
 export function validatePublicKey(pubkeyHex, addressType) {
   if (pubkeyHex === null || pubkeyHex === undefined || pubkeyHex === "") {
@@ -362,7 +362,6 @@ export function validatePublicKey(pubkeyHex, addressType) {
 
   if (
     !isKeyCompressed(pubkeyHex) &&
-    addressType &&
     [P2SH_P2WSH, P2WSH].includes(addressType)
   ) {
     return `${addressType} does not support uncompressed public keys.`;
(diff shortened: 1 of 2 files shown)


src/keys.test.js
```diff
@@ -122,34 +125,73 @@ describe("keys", () => {
...
+      expect(validatePublicKey(uncompressedPublicKey, P2SH)).toBe("");
+    });
+
+    it("returns an empty string when the value is a valid uncompressed public key", () => {
+      expect(validatePublicKey(uncompressedPublicKey, P2SH_P2WSH)).toBe(
+        "P2SH-P2WSH does not support uncompressed public keys."
+      );
+    });
+
+    it("returns an empty string when the value is a valid uncompressed public key", () => {
```
bucko13, line 181: This statement doesn't match the test. should be "returns an error message" Same for the previous test it looks like.

src/keys.js
```diff
@@ -354,6 +360,14 @@ export function validatePublicKey(pubkeyHex) {
     return "Invalid public key.";
   }
 
+  if (
+    !isKeyCompressed(pubkeyHex) &&
+    addressType &&
```
bucko13, line 365: I don't think you need the check for `addressType` since the next line will return false if `addressType` is undefined

src/keys.js
```diff
@@ -330,15 +332,19 @@ export function validateExtendedPublicKey(xpubString, network) {
  * - Must be a valid BIP32 public key.
  *
  * @param {string} pubkeyHex - (compressed) public key in hex
+ * @param {string} addressType - one of P2SH, P2SH-P2WSH, P2WSH
```
bucko13, line 335: needs notation that this is optional.

src/keys.js
```diff
@@ -330,15 +332,19 @@ export function validateExtendedPublicKey(xpubString, network) {
...
  * @returns {string} empty if valid or corresponding validation message if not
  * @example
  * import {validatePublicKey} from "unchained-bitcoin";
  * console.log(validatePublicKey("")); // "Public key cannot be blank."
  * console.log(validatePublicKey("zzzz")); // "Invalid hex..."
  * console.log(validatePublicKey("deadbeef")); // "Invalid public key."
  * console.log(validatePublicKey("03b32dc780fba98db25b4b72cf2b69da228f5e10ca6aa8f46eabe7f9fe22c994ee")); // ""
+ * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53")); // ""
+ * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53", "P2SH")); // ""
+ * console.log(validatePublicKey("04a17f3ad2ecde2fff2abd1b9ca77f35d5449a3b50a8b2dc9a0b5432d6596afd01ee884006f7e7191f430c7881626b95ae1bcacf9b54d7073519673edaea71ee53", "P2WSH")); // "P2WSH doesnt support uncompressed public key."
```
bucko13, line 345: small typo but should be "does not support"

bucko13 (changes requested):
Just a couple small nits to change, but otherwise looks good! Thanks for the update.

src/keys.js
bruce1971: true!

bruce1971:
Thank for reviewing @bucko13, much appreciated!
I added the few minor changes you requested.

If you are fine with all my changes, we could maybe merge into the master branch.
Once merged, ill go over to the caravan repo and make your requested changes there too 👍

src/keys.js
```diff
@@ -330,15 +332,19 @@ export function validateExtendedPublicKey(xpubString, network) {
  * - Must be a valid BIP32 public key.
  *
  * @param {string} pubkeyHex - (compressed) public key in hex
+ * @param {string} addressType - (optional) one of P2SH, P2SH-P2WSH, P2WSH
```
bucko13, line 335: Sorry, I should've been more explicit. There's a specific notation in jsdocs (which is the formatting being used here) for indicating a param is optional (see [docs](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values)). The reason this is important is because we use these to generate [documentation](https://unchained-capital.github.io/unchained-bitcoin/) which is generated based on these comments.  

Here's the sample from the [jsdoc documentation](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values):
```
@param {string} [somebody] - Somebody's name.
```

bucko13 (changes requested):
Sorry, one last nit!

src/keys.js
bruce1971: fixed @bucko13! Thank for pointing it out

waldenraines:
@bucko13 anything else before we merge and release?

bruce1971:
All good from my side!
Thank you for help @bucko13

dhruvbansal (approved):
Thanks for the PR!

bucko13 (approved):

bruce1971:
thx, guys for merging, next week ill try to get related the [caravan PR](https://github.com/unchained-capital/caravan/pull/205) ready 👍

PR #26, opened 2021-01-22
Initial setup introducing typescript support

bucko13:
added dev dependencies and configurations to be able to gradually add typescript notations to the library. Started with `keys.js` -> `keys.ts` to test setup and ease of working in ts/js in parallel.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,6 +19,8 @@
         "sourceType": "module"
     },
     "rules": {
+        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/explicit-module-boundary-types": "off",
         "accessor-pairs": "error",
         "array-bracket-newline": "off",
         "array-bracket-spacing": [
@@ -100,7 +102,7 @@
         ],
         "lines-around-comment": "error",
         "lines-around-directive": "error",
-        "lines-between-class-members": "error",
+        "lines-between-class-members": ["error", "always", {"exceptAfterSingleLine": true}],
         "max-classes-per-file": "error",
         "max-depth": "error",
         "max-len": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,11 +5,15 @@
         "node": true,
         "jest/globals": true
     },
-    "extends": "eslint:recommended",
+    "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended"
+    ],
     "globals": {
         "Atomics": "readonly",
         "SharedArrayBuffer": "readonly"
     },
+    "parser": "@typescript-eslint/parser",
     "parserOptions": {
         "ecmaVersion": 2018,
         "sourceType": "module"
@@ -304,5 +308,5 @@
             "never"
         ]
     },
-    "plugins": ["jest"]
+    "plugins": ["jest", "@typescript-eslint"]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -68,7 +68,7 @@
             "property"
         ],
         "dot-notation": "error",
-        "eol-last": "off",
+        "eol-last": "error",
         "eqeqeq": "off",
         "func-call-spacing": "error",
         "func-name-matching": "error",
--- a/src/keys.ts
+++ b/src/keys.ts
@@ -16,6 +16,7 @@ import { bip32PathToSequence, validateBIP32Path } from "./paths";
 import { TESTNET, networkData, MAINNET, NETWORKS } from "./networks";
 import { P2SH_P2WSH } from "./p2sh_p2wsh";
 import { P2WSH } from "./p2wsh";
+import { KeyPrefix, KeyVersion } from "./types";
 
 export const EXTENDED_PUBLIC_KEY_VERSIONS = {
   xpub: "0488b21e",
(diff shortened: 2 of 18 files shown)


bucko13:
Deliberately left a type error in commit d84b199 to see how the CI responded (note the x). I would personally be in favor of letting these error out as they will only apply to files that have been explicitly set to `.ts` extension, but if people prefer we can set them to warnings.

dhruvbansal (dismissed):

bucko13:
One thing this PR doesn't do is to have a command that emits the declaration files (--emitDeclarationOnly from the [compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) which would let other typescript enabled "understand" and leverage the types as we introduce them by exposing them in the final build. Not sure if this is something we want to introduce later when we've got broader support or if we'd rather add it in now, so throwing out the idea here.

dhruvbansal:
> One thing this PR doesn't do is to have a command that emits the declaration files (--emitDeclarationOnly from the [compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) which would let other typescript enabled "understand" and leverage the types as we introduce them by exposing them in the final build. Not sure if this is something we want to introduce later when we've got broader support or if we'd rather add it in now, so throwing out the idea here.

I think we should enable it.  Next we can move unchained-wallets over to TS and try it out.

bucko13:
Alright, the build will now emit type declaration files for everything (only `.ts` files will emit something of any meaning) which makes unchained-bitcoin properly ts compatible for any ts projects that want to use it.

bucko13:
The linter was picking up that it didn't like the `require` (commonjs style) imports for `bitcoinjs` module and so all those files needed to be updated. the default imports also didn't work when compiled with our babel settings so each of those files needed to be specific about what they wanted from the module (which is a better pattern anyway). The unfortunate thing is it made the PR a little bigger though.

waldenraines:
@dhruvbansal anything further?

.eslintrc.json
```diff
@@ -304,5 +310,5 @@
             "never"
         ]
     },
-    "plugins": ["jest"]
+    "plugins": ["jest", "@typescript-eslint"]
 }
```
humanumbrella, line 314: missing eof line

tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    // Ensure that Babel can safely transpile files in the TypeScript project
+    "isolatedModules": true,
+    "outDir": "lib",
+	},
+	"include": [
+		"src"
+	]
+}
```
humanumbrella, line 24: eof

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
humanumbrella, line 1: tab spacing is weird in here

src/keys.ts
```diff
@@ -28,15 +28,18 @@ export const EXTENDED_PUBLIC_KEY_VERSIONS = {
   vpub: "045f1cf6",
   Upub: "024289ef",
   Vpub: "02575483",
-};
+} as const;
+
+type KeyPrefixes = keyof typeof EXTENDED_PUBLIC_KEY_VERSIONS;
```
humanumbrella, line 33: Is this only required for the next line? Or could this be used in `validatePrefix` ?

src/keys.ts
```diff
@@ -72,24 +75,46 @@ export function validateRootFingerprint(rootFingerprint) {
...
+      this.sequence = bip32PathToSequence(this.path);
+      this.index = this.sequence[this.sequence.length - 1];
+      this.depth = this.path.split("/").length - 1;
+    } else {
+      assert(
+        options.depth && options.index,
+        'Either an absolute bip32 path or index and depth are required to create ExtendedPublicKey'
+      )
+      this.depth = options.depth;
+      this.index = options.index;
```
humanumbrella, line 108: how bout a test to keep us at 100 coverage

humanumbrella (changes requested):
So yeah -- this PR got a bit bigger in scope when fixing the linting errors for the bitcoin-js imports. That's okay I suppose, but it does impact testing scope. 

Screen Shot 2021-02-16 at 11 36 45 PM


I don't see anything wrong with the changes. Only a couple nits a question or two. Should be able to approve tomorrow I'd say.

src/keys.ts
bucko13: good call!

src/keys.ts
bucko13: haha. good catch!

tsconfig.json
bucko13: another instance of my editor hiding stuff form my apparently. Looks like tabs were being used for tabs instead of spaces. I hope I got it fixed but i guess I won't know until I push up 😬

ReyesSalinas (commented):
from a typing standpoint this LGTM

src/keys.ts
```diff
@@ -28,15 +28,18 @@ export const EXTENDED_PUBLIC_KEY_VERSIONS = {
   vpub: "045f1cf6",
   Upub: "024289ef",
   Vpub: "02575483",
-};
+} as const;
+
+type KeyPrefix = keyof typeof EXTENDED_PUBLIC_KEY_VERSIONS;
+type KeyVersions = typeof EXTENDED_PUBLIC_KEY_VERSIONS[KeyPrefix];
```
ReyesSalinas, line 34: Would prefer these moved to another file to help clean up this already big file

ReyesSalinas (approved):
approved from my end

humanumbrella (approved):
LGTM

PR #29, opened 2021-03-04
export psbt magic

bucko13:
Since we're needing this data in other places, it seemed useful to export as a constant. It also helps to self-document when it's being used rather than just look like a random hex string.

### DIFF
--- a/src/psbt.js
+++ b/src/psbt.js
@@ -49,6 +49,8 @@ import {generateBip32DerivationByIndex, generateBraid} from './braid';
  *
  */
 
+export const PSBT_MAGIC_HEX = "70736274ff";
+export const PSBT_MAGIC_B64 = "cHNidP8"; 
 
 /**
  * Return the getBip32Derivation (if known) for a given `Multisig` object.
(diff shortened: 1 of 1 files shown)


src/psbt.js
```diff
@@ -49,6 +49,8 @@ import {generateBip32DerivationByIndex, generateBraid} from './braid';
  *
  */
 
+export const PSBT_MAGIC_HEX = "70736274ff";
+export const PSBT_MAGIC_B64 = "cHNidP";
```
humanumbrella, line 53: I probably didn't actually compute this back then, but it should be 

`cHNidP8=` as a conversion of 0x70736274FF to base64

humanumbrella (changes requested):

src/psbt.js
bucko13: Yeah, I saw that too but didn't want to rock the boat too much. In fact, I think it needs to be without the `=` though as that's added by base64 as a buffer for the end of a string I think. Anyway, that breaks the tests but adding an `8` at the end to finish the conversion seems to work (and then lower down `substring(0,7)`

bucko13: That's what it looks like from the [test vectors](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#test-vectors) too.

humanumbrella (approved):

PR #37, opened 2021-08-18
Add PSBT translation for non-PSBT devices

humanumbrella:
NOTE:  currently only working for `P2SH` addresses.

The idea is to take a normal unsigned PSBT and translate it into something consumable by non-PSBT devices (e.g. a Ledger) This functionality can then be taken advantage of from within the [unchained-wallets](https://github.com/unchained-capital/unchained-wallets/) library.

I have added functionality to

1. Take an unsigned PSBT and translate it so as to return `inputs`, `outputs`, and `bip32paths` which is what's expected to be provided to other devices
2. Given an unsigned PSBT and arrays of public keys and signatures, insert those signature(s) into the PSBT on the proper inputs

Let's discuss.


src/psbt.js
```diff
@@ -252,16 +384,13 @@ export function psbtOutputFormatter(output) {
...
-    psbt = Psbt.fromHex(psbtFromFile);
-  } else if (psbtFromFile.substring(0, 7) === PSBT_MAGIC_B64) {
-    psbt = Psbt.fromBase64(psbtFromFile);
-  } else {
-    return null;
-  }
+  // this is only if you are providing a "FULLY SIGNED" PSBT
+  // if (!psbt.validateSignaturesOfAllInputs()) {
+  //   return null;
+  // }
```
waldenraines, line 393: Can remove?

waldenraines (commented):
> NOTE: currently only working for `P2SH` addresses.

Are we planning on supporting other addresses prior to merge?

How can I go about testing this locally?

waldenraines:
See https://github.com/unchained-capital/unchained-wallets/pull/61

src/psbt.js
```diff
@@ -231,6 +234,135 @@ export function psbtOutputFormatter(output) {
...
+  if (psbtWithSignatures === null) return null;
+
+  signatures.forEach((sig, idx) => {
+    const pubkey = pubkeys[idx];
+      psbtWithSignatures = _addSignatureToPSBT(psbtWithSignatures, idx, pubkey, sig);
+  }
+  )
+  return psbtWithSignatures.toBase64();
+}
 /**
```
bucko13, line 366: nit for a space between functions.

src/psbt.test.js
```diff
@@ -1,13 +1,22 @@
 import {
+  addSignaturesToPSBT, autoLoadPSBT,
```
bucko13, line 2: can these be split into two lines just to keep consistent? (grrr, prettier would've done this for you I'm pretty sure)

src/psbt.test.js
```diff
@@ -39,6 +48,110 @@ describe("psbt", () => {
...
...0000000000ffffffff01b3bff5050000000017a91420fd9addb8034a12ce353b8aa7f4faca50259b198700000000000100f702000000000101ba408eed9f0758c35eca8c0ef8e29cb63eed4c4d1b1349c989ce681f4f398f270000000017160014aba867b774169ca1e1bf635c204e3a4a80d7d0f9feffffff0200e1f5050000000017a914f745b90f312a38249a1e05c5db3a2f8d0aca4fdd87a02a29150100000017a9141d8a90a10f4ff666b36e797a4d00df9d36bd213c8702473044022004593ebe490574537724328a73b206863aea72b6b51b677e0aee6f7aa2afe54f0220560bd4a3665e34d559e96f77f6e6b619d214b36df04bcb58ffe0d74174cbd8e601210208741a0d458e76b355d1c274d78d8c165e207be0a5ff50623c0d63eb7932fa2f00000000010469522102315cbe4eada106463e04c238ffa3dbf26a77ad5047ac769c43f4629525266f4521029e866b92b02c686f89b320c0ce4c2391ee531b4dccfd719aefc8caa82e42c8282103ba8cb8627165090c2dca871274760b1ef027d89c6f54eb076e567a4f6c5e850853ae220603ba8cb8627165090c2dca871274760b1ef027d89c6f54eb076e567a4f6c5e85081cefa5d9162d00008001000080000000800100000000000000000000002206029e866b92b02c686f89b320c0ce4c2391ee531b4dccfd719aefc8caa82e42c8281cf57ec65d2d0000800100008032000080000000000000000000000000220602315cbe4eada106463e04c238ffa3dbf26a77ad5047ac769c43f4629525266f451cf57ec65d2d000080010000803c0000800000000000000000000000000000';
```
bucko13, line 157: Obviously this isn't new, but is there a reason these were included in the test file and not fixtures?

src/psbt.js
```diff
@@ -231,6 +234,135 @@ export function psbtOutputFormatter(output) {
...
+  // To provide the non-PSBT wallet with the proper info, we actually
+  // need to grab some info from the "data inputs" and some other
+  // information from the "transaction inputs".
+  let txInputs = localPSBT.txInputs.map((input, index) => {
+    const dataInput = localPSBT.data.inputs[index];
+    // The bip32Derivation hanging off of the dataInput contains
+    // an array of objects like:
+    //    [
+    //      { masterFingerprint1, pubkey1, path1 },
+    //      { masterFingerprint2, pubkey2, path2 },
```
bucko13, line 268: `masterFingerprint` is the same as `rootFingerprint` and `xfp`?

src/psbt.js
```diff
@@ -231,6 +234,135 @@ export function psbtOutputFormatter(output) {
...
+ */
+export function translatePSBT(network, addressType, psbt, keyDetails) {
+  let localPSBT = autoLoadPSBT(psbt, {network: networkData(network)});
+  if (localPSBT === null) return null;
+
+  let bip32Derivations = [];
+  // To provide the non-PSBT wallet with the proper info, we actually
+  // need to grab some info from the "data inputs" and some other
+  // information from the "transaction inputs".
+  let txInputs = localPSBT.txInputs.map((input, index) => {
```
bucko13, line 262: Any way to split this up like the `addSignatureToPSBT` for readability?

src/psbt.js
```diff
@@ -231,6 +234,135 @@ export function psbtOutputFormatter(output) {
...
+ * Translates a PSBT into inputs/outputs consumable by non-PSBT devices in the
+ * `unchained-wallets` library.
+ *
+ * FIXME - Have only confirmed this is working for P2SH addresses on Ledger on regtest
+ *
+ * @param {module:networks.NETWORKS} network - bitcoin network
+ * @param {module:multisig.MULTISIG_ADDRESS_TYPES} addressType - address type
+ * @param {String} psbt - PSBT as a base64 or hex string
+ * @param {Object} keyDetails - Object containing signing key details (Fingerprint + bip32 prefix)
+ * @returns {null|Object} returns object with the format
```
bucko13, line 247: If we were to name the return object what would it be? We have a lot of these different objects with different shapes that are returned in this library, and I'm wondering if there is any other corollary. While we're not going to do it in this PR, Typescript helps here because it forces us to be explicit about the _type_ that is returned and then we can easily keep track (and let our IDEs do so as well)

src/psbt.js
```diff
@@ -231,6 +234,135 @@ export function psbtOutputFormatter(output) {
...
+/**
+ * Translates a PSBT into inputs/outputs consumable by non-PSBT devices in the
+ * `unchained-wallets` library.
+ *
+ * FIXME - Have only confirmed this is working for P2SH addresses on Ledger on regtest
+ *
+ * @param {module:networks.NETWORKS} network - bitcoin network
+ * @param {module:multisig.MULTISIG_ADDRESS_TYPES} addressType - address type
+ * @param {String} psbt - PSBT as a base64 or hex string
+ * @param {Object} keyDetails - Object containing signing key details (Fingerprint + bip32 prefix)
```
bucko13, line 246: Are key details the same as "ExtendedPublicKeyDerivation" (or maybe `KeyDerivation` to be more generic)? 

In past discussions (and in the bitcoinjs library from which this is based) we've defined this as something like:

```
interface PublicKeyDerivation {
  rootFingerprint: Buffer | string;
  pubkey: Buffer | string; 
  path: string;
}
```

the xpub derivation being the same except `xpub` instead of `pubkey`. Maybe the base "class" could be:

```
interface KeyDerivation {
  rootFingerprint: Buffer | string // alternatively could be xfp?
  path: string
}
```

bucko13 (changes requested):
Looking good. Caught some errors from the linter and had some other comments about formatting. I also have some questions about naming. Since these objects and the way they are shaped are depended on across libraries, how they're named and what their shape is important b/c it'll be hard to change so I think it's worth making sure it's settled as much as possible (keyDetails and its props was the one I had the most comments on) 

Running `npm run lint` caught 7 errors:

```
unchained-bitcoin/src/psbt.js
  275:69  error  Expected to return a value at the end of arrow function  array-callback-return
  275:69  error  Expected to return a value at the end of arrow function  consistent-return
  324:1   error  Unexpected dangling '_' in '_addSignatureToPSBT'         no-underscore-dangle

unchained-bitcoin/src/psbt.test.js
  74:22  error  Unexpected block statement surrounding arrow body; parenthesize the returned value and move it immediately after the `=>`  arrow-body-style
  76:27  error  Missing radix parameter                                                                                                    radix
  94:21  error  Unexpected block statement surrounding arrow body; parenthesize the returned value and move it immediately after the `=>`  arrow-body-style

unchained-bitcoin/src/script.js
  44:29  error  'script' is already declared in the upper scope  no-shadow
```

src/psbt.test.js
humanumbrella: IIRC -- bc they're not related to the key material in the fixtures

src/psbt.js
humanumbrella: Yes

src/psbt.js
```diff
@@ -292,3 +421,21 @@ export function parseSignaturesFromPSBT(psbtFromFile) {
...
   return signatureSet;
 }
+
+/**
+ * Given a string, try to create a Psbt object based on MAGIC (hex or Base64)
+ * @param {String} psbtFromFile -  Base64 or hex PSBT
+ * @param {Object} [options] -  options, e.g. TESTNET
+ * @return {null|Psbt}
+ */
+export function autoLoadPSBT(psbtFromFile, options=null) {
```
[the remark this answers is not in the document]
humanumbrella: no reason, and moved it up

src/psbt.js
```diff
@@ -231,6 +234,135 @@ export function psbtOutputFormatter(output) {
...
+  ]
+  psbt.data.updateInput(inputIndex, {partialSig})
+  if (!psbt.validateSignaturesOfInput(inputIndex, pubkey))
+    throw new Error("One or more invalid signatures.");
+  return psbt;
+}
+
+/**
+ * Given an unsigned PSBT, an array of signing public key(s) (one per input),
+ * an array of signature(s) (one per input) in the same order as the pubkey(s),
```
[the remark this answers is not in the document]
humanumbrella: Agree it's somewhat brittle. That said -- this isn't really how signing is supposed to work on non-psbt devices. The data used here actually comes from two different places. We actually create the public key array in the constructor of the interaction class because we have the info at that time (in the derivations) - it is created in the order of the inputs as we're looping through them then.

The signatures that come back from the device as an array are also in the order of the inputs - the device itself doesn't send back public keys after signing and we can't really control what it sends back to us.

The memory of order assumption hell comes about where we are doing a sort of inverse of what's happening here -- e.g. a partially- or fully-signed PSBT comes back from a device and instead of just **using** that psbt, we want to pull the signatures out of it to create a pubkey keyed dictionary of signature arrays (... which must be ordered (properly!) to work given the signature validation scheme) to match what the consuming application expected. More refactoring/dropping of that kind of code likely in the future.

src/psbt.js
humanumbrella: hmm - I don't know what I'd call it. `txIngredients` ? `txDetails` ? `txInfo`? `signingDetails`? We're not really defining anything new here, it's just hey - this is what we already know the devices will accept.

It's all the stuff the nonPSBT device needs to sign the transaction (minus the network).

src/psbt.js
humanumbrella: No, not really. It happens to be compatible with the `KeyDerivation` you defined, but in this case it's not really what we normally mean. e.g. the path we're providing is not the path to the public key but rather the root path.

Here's the problem:

In the general case we should only need to provide the `rootFingerprint` || `masterFingerprint` || `xfp` to know which device is trying to sign, HOWEVER -- in the case where you put multiple keys from the same device in the same multisig ... aka lots of test cases everywhere ... we need to be able to differentiate who is tryting to sign and this is a way to do that.

src/psbt.js
humanumbrella: open to suggestions - I tried to do a couple things but ended up reverting back to this. Getting the necessary pieces from the right place is a little bit tangled. I think the comments help but ofc I am biased.

src/psbt.js
bucko13: I'm not totally clear on what `rootPath` is? The first thing that comes to mind is the path of the root key, but that would just be `m` of course and irrelevant/redundant. Next is the path from which the xpub is derived. But isn't that just the path that's in the `ExtendedPublicKeyDerivation`?

```
interface ExtendedPublicKeyDerivation {
    rootFingerprint: Buffer | string;
    xpub: string;
    path: string;
}
```
In the [descriptor spec](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md#key-origin-identification), this grouping is called the `KeyOrigin`. To add to the confusion, in the bip174/bitcoinjs library this is `Bip32Derivation` ([source](https://github.com/bitcoinjs/bip174/blob/master/ts_src/lib/interfaces.ts#L85))

```
export interface Bip32Derivation {
  masterFingerprint: Buffer;
  pubkey: Buffer;
  path: string;
}
```

Maybe I'm just misunderstanding something? Or another way to put it is you're just trying to make a distinction between a path to a public key that's used to generate an address and the path used to _find_ the origin of the xpub? In which case, I think we're saying the same thing. Basically there's a derivation for a pubkey which would have a path to get that, and a derivation for an xpub which would be used to get that extended pubkey. 

My main objection is just adding more new vocabulary, especially in proptypes where that's going to be important, especially if it's not in what's being used elsewhere. This has already been a really confusing aspect of working with this library in the past. 

The name of the variable isn't ultimately too important b/c it can be changed, though my preference is to stick to some of what we've already talked about and have been using, which is above. For the properties, I think it should just be `path` or `bip32Path`. `root` feels very confusing to me and isn't used in other places i've seen, and in the context of an xpub derivation, the path could only really refer to the "origin" of that xpub (as it's used in descriptors).

humanumbrella: Sure, can rename it - hopefully makes it clearer.

src/psbt.js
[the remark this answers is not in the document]
bucko13: Thanks for the recall! That makes sense.

src/psbt.js
bucko13: A rough draft of something that might work: https://github.com/humanumbrella/unchained-bitcoin/pull/2. Kind of related to the previous comment about the `txDetails` (?) below. breaking these out into more functional pieces might help. There are some loops that could be consolidated, but that's not unusual in functional stuff from what I've seen (e.g. chained things like `arr.filter(...).map(...)`.

src/psbt.js
humanumbrella: going to go back on what i said here a bit just to say that it's the unchained-specific object that does contain txDetails but in a special way that really only we care about (especially right now)

src/psbt.js
humanumbrella: closed that one and edited it a bit - lmk

bucko13:
I think this update to make it a bit more functional, which is consistent with the philosophy of the library, is a lot better and much more readable. 

The idea that certain functions/return objects are named to be unchained specific does feel a little problematic. I think this is more a reflection of a problem that we have throughout the library and so would be harder to change. If I had to put my finger on why this is a problem, I think it's because there are places where when certain properties are needed, they are just added to an object and then passed around. This gets us into a place where we need these very specially shaped object to do certain actions. 

An example is the `unchainedInputs` that we now have a function to return. A `PsbtTxInput` in bitcoinjs looks like this:

```
export interface PsbtTxInput {
    hash: string | Buffer;
    index: number;
    sequence?: number;
    hash: Buffer;
}
```

which is very similar to the unchained inputs except for another object we add in as a property (this is extrapolating an interface based on the code here):

```
interface UnchainedInput { 
      amountSats: number
      index: number
      transactionHex: string
      txid: string
      multisig: Multisig
}
```

where the type Multisig is:

```
interface Multisig {
  address: string | Address
  redeem: Payment // another object w/ 14 of its own properties, from bitcoinjs
  multisigBraidDetails: BraidDetails // another object defined in unchained libs
  bip32Derivations: Bip32Derivations[] // an array of more objects defined in unchained libs
}

```

So this object being returned has items that are three levels deep, and hard to define or know about if you're not really familiar with these objects.

This could be solved in two ways in my opinion (both of which get aided by Typescript which we now have in the library):

1. using classes where the properties are clearly defined, and methods attached to help get some of the harder to reach elements.
2. Smaller, more fine-tuned functions, where, rather than having functions that return complicated objects, we would be more forced to compose functions to get things out that we need for certain use cases. So, rather than having the multisig object in the `unchainedInputs` object because we might need the braid details, we would just have another method `getBraidDetailsFromPSBT` (or whatever) that would be called when that info is required. 

Obviously the second fits more in line with the philosophy of this project. I think the latest change also brings this branch more in line with this philosophy which is good!

I wouldn't call these comments blockers by any means since it's the result of dealing with things that are already baked in and changing them would be out of scope. I would argue for not even having the `translatePSBT` function at all since that's just another instance of returning somewhat arbitrary objects, and instead using the functions directly in uc-wallets where you need this work done. But again, doesn't need to be a blocker. 

I'm curious if @dhruvbansal has any thoughts in particular about the naming now that we have these very explicitly "Unchained" objects.

src/psbt.js
```diff
@@ -231,6 +252,163 @@ export function psbtOutputFormatter(output) {
...
+  });
+}
+
+/**
+ * Create unchained-wallets style transaction output objects from a PSBT
+ *
+ * @param {Object} psbt - Psbt bitcoinjs-lib object
+ * @return {Object[]} unchained multisig transaction outputs array
+ */
+function getUnchainedOutputsFromPSBT(psbt) {
```
bucko13, line 288: Do you think this could just be `getOutputsFromPSBT`? This is very close to the outputs in bitcoinjs too, just renaming the properties slightly, which is fine. I would say that we could even return with the `script` property since [that's in these objects to begin with.](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/types/psbt.d.ts#L15) and then just call them outputs.

bucko13 (commented):
My preference would be to use the new functions directly without the translate function at all, and also I think we could make the suggested update to the outputs function and change it to `getOutputsFromPsbt`. I won't block on those though if there's any disagreement. 

I would like to get @dhruvbansal 's feedback if possible before merging, specifically having these be explicitly "unchained" functions. 

Barring the above, this looks good.

dhruvbansal (approved):

waldenraines:
@bucko13 @humanumbrella was a resolution reached on [this comment](https://github.com/unchained-capital/unchained-bitcoin/pull/37#pullrequestreview-737682598)?

bucko13:
> was a resolution reached on this comment?

My comment was that if dhruv approves then I'm ok punting.

bucko13 (approved):

PR #41, opened 2021-11-13
fix(package.json): add bip66

KayBeSee:
adds bip66 to package.json which is referenced in src/signatures.js


KayBeSee:
Yarn2 throws an error when building packages because `bip66` is used in `src/signatures.js` but not declared in the `package.json`.

bucko13 (approved):
Looks good to me. I guess this was working before because it's an upstream dependency of something else we're already using, but making it explicit is good. Thanks!

PR #56, opened 2023-02-09
fix: remove module from typedef interfering with d.ts exports

bucko13:
I found that this project was having trouble getting built in typescript files (namely unchained-wallets in this case) because it would run into some namespace issues with declaration files that were generated from the jsdoc documentation. In particular it didn't like the `typedef module:...` because it would create a type of `module` that it would export from the module. 

I have to test still if this actually breaks some of the documentation generation (at worst it's probably just the linking). But I've confirmed it does allow it to get built in unchained-wallets when converting that library to typescript


Shadouts (approved):
Seems right, but I'm not familiar with jsdoc. :shrug:

humanumbrella (approved):

PR #59, opened 2023-02-14
PsbtV2 Class

Shadouts:
Just putting this up as a draft for now to get some eyes on it. Currently working on unit tests and additional validation. If other stuff is ready to use this, maybe we can merge now for exploratory testing.


Shadouts:
Run the following as a script to get an idea of this in action:
```
const lib = require('./lib');
const psbt = lib.PsbtV2.DeserializeFromV0("70736274ff0100750200000001268171371edff285e937adeea4b37b78000c0566cbb3ad64641713ca42171bf60000000000feffffff02d3dff505000000001976a914d0c59903c5bac2868760e90fd521a4665aa7652088ac00e1f5050000000017a9143545e6e33b832c47050f24d3eeb93c9c03948bc787b32e1300000100fda5010100000000010289a3c71eab4d20e0371bbba4cc698fa295c9463afa2e397f8533ccb62f9567e50100000017160014be18d152a9b012039daf3da7de4f53349eecb985ffffffff86f8aa43a71dff1448893a530a7237ef6b4608bbb2dd2d0171e63aec6a4890b40100000017160014fe3e9ef1a745e974d902c4355943abcb34bd5353ffffffff0200c2eb0b000000001976a91485cff1097fd9e008bb34af709c62197b38978a4888ac72fef84e2c00000017a914339725ba21efd62ac753a9bcd067d6c7a6a39d05870247304402202712be22e0270f394f568311dc7ca9a68970b8025fdd3b240229f07f8a5f3a240220018b38d7dcd314e734c9276bd6fb40f673325bc4baa144c800d2f2f02db2765c012103d2e15674941bad4a996372cb87e1856d3652606d98562fe39c5e9e7e413f210502483045022100d12b852d85dcd961d2f5f4ab660654df6eedcc794c0c33ce5cc309ffb5fce58d022067338a8e0e1725c197fb1a88af59f51e44e4255b20167c8684031c05d1f2592a01210223b72beef0965d10be0778efecd61fcac6f79a4ea169393380734464f84f2ab300000000000000");

console.log('-- globals --')
console.log('PSBT_GLOBAL_XPUB:', psbt.PSBT_GLOBAL_XPUB)
console.log('PSBT_GLOBAL_TX_VERSION:', psbt.PSBT_GLOBAL_TX_VERSION)
console.log('PSBT_GLOBAL_FALLBACK_LOCKTIME:', psbt.PSBT_GLOBAL_FALLBACK_LOCKTIME)
console.log('PSBT_GLOBAL_INPUT_COUNT:', psbt.PSBT_GLOBAL_INPUT_COUNT)
console.log('PSBT_GLOBAL_OUTPUT_COUNT:', psbt.PSBT_GLOBAL_OUTPUT_COUNT)
console.log('PSBT_GLOBAL_TX_MODIFIABLE:', psbt.PSBT_GLOBAL_TX_MODIFIABLE)
console.log('PSBT_GLOBAL_VERSION:', psbt.PSBT_GLOBAL_VERSION)
console.log('PSBT_GLOBAL_PROPRIETARY:', psbt.PSBT_GLOBAL_PROPRIETARY)

console.log('-- inputs --')
console.log('PSBT_IN_NON_WITNESS_UTXO:', psbt.PSBT_IN_NON_WITNESS_UTXO)
console.log('PSBT_IN_WITNESS_UTXO:', psbt.PSBT_IN_WITNESS_UTXO)
console.log('PSBT_IN_PARTIAL_SIG:', psbt.PSBT_IN_PARTIAL_SIG)
console.log('PSBT_IN_SIGHASH_TYPE:', psbt.PSBT_IN_SIGHASH_TYPE)
console.log('PSBT_IN_REDEEM_SCRIPT:', psbt.PSBT_IN_REDEEM_SCRIPT)
console.log('PSBT_IN_WITNESS_SCRIPT:', psbt.PSBT_IN_WITNESS_SCRIPT)
console.log('PSBT_IN_BIP32_DERIVATION:', psbt.PSBT_IN_BIP32_DERIVATION)
console.log('PSBT_IN_FINAL_SCRIPTSIG:', psbt.PSBT_IN_FINAL_SCRIPTSIG)
console.log('PSBT_IN_FINAL_SCRIPTWITNESS:', psbt.PSBT_IN_FINAL_SCRIPTWITNESS)
console.log('PSBT_IN_POR_COMMITMENT:', psbt.PSBT_IN_POR_COMMITMENT)
console.log('PSBT_IN_RIPEMD160:', psbt.PSBT_IN_RIPEMD160)
console.log('PSBT_IN_SHA256:', psbt.PSBT_IN_SHA256)
console.log('PSBT_IN_HASH160:', psbt.PSBT_IN_HASH160)
console.log('PSBT_IN_HASH256:', psbt.PSBT_IN_HASH256)
console.log('PSBT_IN_PREVIOUS_TXID:', psbt.PSBT_IN_PREVIOUS_TXID)
console.log('PSBT_IN_OUTPUT_INDEX:', psbt.PSBT_IN_OUTPUT_INDEX)
console.log('PSBT_IN_SEQUENCE:', psbt.PSBT_IN_SEQUENCE)
console.log('PSBT_IN_REQUIRED_TIME_LOCKTIME:', psbt.PSBT_IN_REQUIRED_TIME_LOCKTIME)
console.log('PSBT_IN_REQUIRED_HEIGHT_LOCKTIME:', psbt.PSBT_IN_REQUIRED_HEIGHT_LOCKTIME)
console.log('PSBT_IN_TAP_KEY_SIG:', psbt.PSBT_IN_TAP_KEY_SIG)
console.log('PSBT_IN_TAP_SCRIPT_SIG:', psbt.PSBT_IN_TAP_SCRIPT_SIG)
console.log('PSBT_IN_TAP_LEAF_SCRIPT:', psbt.PSBT_IN_TAP_LEAF_SCRIPT)
console.log('PSBT_IN_TAP_BIP32_DERIVATION:', psbt.PSBT_IN_TAP_BIP32_DERIVATION)
console.log('PSBT_IN_TAP_INTERNAL_KEY:', psbt.PSBT_IN_TAP_INTERNAL_KEY)
console.log('PSBT_IN_TAP_MERKLE_ROOT:', psbt.PSBT_IN_TAP_MERKLE_ROOT)
console.log('PSBT_IN_PROPRIETARY:', psbt.PSBT_IN_PROPRIETARY)

console.log('-- outputs --')
console.log('PSBT_OUT_REDEEM_SCRIPT:', psbt.PSBT_OUT_REDEEM_SCRIPT)
console.log('PSBT_OUT_WITNESS_SCRIPT:', psbt.PSBT_OUT_WITNESS_SCRIPT)
console.log('PSBT_OUT_BIP32_DERIVATION:', psbt.PSBT_OUT_BIP32_DERIVATION)
console.log('PSBT_OUT_AMOUNT:', psbt.PSBT_OUT_AMOUNT)
console.log('PSBT_OUT_SCRIPT:', psbt.PSBT_OUT_SCRIPT)
console.log('PSBT_OUT_TAP_INTERNAL_KEY:', psbt.PSBT_OUT_TAP_INTERNAL_KEY)
console.log('PSBT_OUT_TAP_TREE:', psbt.PSBT_OUT_TAP_TREE)
console.log('PSBT_OUT_TAP_BIP32_DERIVATION:', psbt.PSBT_OUT_TAP_BIP32_DERIVATION)
console.log('PSBT_OUT_PROPRIETARY:', psbt.PSBT_OUT_PROPRIETARY)

console.log('-- serialized --')
console.log(psbt.serialize('hex'))
```

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+        globalXpub.extendedPubkey,
+        globalXpub.masterFingerprint,
+        globalXpub.path,
+      );
+    }
+
+    let txInputs:any = [];
+    for (const [index, txInput] of psbtv0.txInputs.entries()) {
+      txInputs[index] = txInput;
+    }
```
tylerchambers, line 875: could we make this a const and use a map operation?

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+        witnessScript: input.witnessScript,
+        bip32Derivation: input.bip32Derivation,
+      });
+    }
+
+    let txOutputs:any = [];
+    for (const [index, txOutput] of psbtv0.txOutputs.entries()) {
+      txOutputs[index] = txOutput;
+    }
+
```
tylerchambers, line 895: same as above

tylerchambers (changes requested):
Nice! A few pieces of feedback. More generally, too, it would be nice to add a space after the colon when you're doing typing as that seems to be the more popular convention. other than that, great stuff

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+    (/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{4})$/).test(psbt)
+  ) {
+    // Is a base64 string
+    buf = Buffer.from(psbt, 'base64');
+  } else {
+    throw ERROR_PSBT_NOT_VALID;
+  }
+
+  return buf;
+}
```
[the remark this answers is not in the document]
Shadouts: Perhaps this could use some better structure, but I don't know that everything in your suggestion is better. The base64 check I committed does a more robust check of valid formats. During development, I was running into cases where it was switching to the wrong format case. The regex, though long, ensures valid padding and chunking of base64. Likewise, the hex regex is checking for full bytes. If a byte is a nibble nothing will work, so it with throw.

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+    const values:({ key:string, value:string|null})[][] = [];
+
+    for (const map of maps) {
+      values.push(
+        // There is probably a better way to type this
+        getNonUniqueKeyTypeValues(map, keytype) as { key:string, value:string|null }[]
+      );
+    }
+
+    return values;
```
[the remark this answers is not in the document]
Shadouts: I can get on board with the map method, but def not the flat. It's important that the inputs/outputs retain their indexing.

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
+import { BufferReader, BufferWriter } from 'bufio';
```
bucko13, line 1: Might be nice to link to the BIP for reference at the top of the file: https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki

bucko13: And then for the various consts and enums that are dictated by the BIP itself we can similarly indicate that they should match the respective section of the BIP

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+const ERROR_PSBT_NOT_VALID = Error("Not a valid psbt");
+
+
+/* 
+Helper Functions
+*/
+
+// Return psbt as Buffer.
+function bufferize (psbt:string|Buffer):Buffer {
+  let buf:Buffer|undefined;
```
bucko13, line 98: I think I see where some of the frustration is coming from with the quote formatting. We don't have prettier setup in anyway in this repo so there's nothing to run a "format on save" with. Just noticing it here because prettier would add spacing here:

```suggestion
  let buf: Buffer | undefined;
```

bucko13: I think it'll just be easier to add a pre-commit hook setup. I'll do this in another PR if there aren't any objections. Might also add commit lint (like we have in caravan and elsewhere) which should make auto deploys easier in the future if we want.

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+
+  if (Buffer.isBuffer(psbt)) {
+    buf = psbt;
+  } else if (typeof psbt === 'string' && (/^([0-9A-Fa-f][0-9A-Fa-f])+$/).test(psbt)) {
```
bucko13, line 102: heads up we do have a [`validateHex` method](https://github.com/unchained-capital/unchained-bitcoin/blob/a6ca2ac8a1fe6be50476455cf8a930e681f888d7/src/utils.js#L49-L58)

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+    // Is a hex string
+    buf = Buffer.from(psbt, 'hex');
+  } else if (
+    typeof psbt === 'string' && 
+    (/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{4})$/).test(psbt)
+  ) {
+    // Is a base64 string
```
bucko13, line 109: turns out we don't have a tool for this but might be nice to have and then the code can be self-documenting. e.g. `else if (isBase64(psbt))`.

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+        // There is probably a better way to type this
+        getNonUniqueKeyTypeValues(map, keytype) as { key:string, value:string|null }[]
+      );
+    }
+
+    return values;
+  }
+  
+  const map = maps;
+  const vals:{ key:string, value:string|null }[] = [];
```
bucko13, line 144: this type is being repeated. will think more on if there's a better way to do the typing above but at the very least this could be pulled out as something like 

```typescript
type UniqueKeyTypeVals = { key:string, value:string|null }[]
```

or something

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+    if (node.indexOf('\'') > -1) {
+      // Hardened node needs hardening
+      num += BIP_32_HARDENING_OFFSET;
+    }
+
+    bw.writeU32(num);
+  }
+
+  return bw.render();
+}
```
bucko13, line 201: Just noting that we have another module in this library that might have some useful utilities for this, at the very least to reuse the regexes we already have. Checkout `paths.js`

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+  OUTPUTS = 'OUTPUTS', // 0b000000010
+  SIGHASH_SINGLE = 'SIGHASH_SINGLE', // 0b000000100
+}
+
+
+/* 
+Global Constants
+ */
+
+const PSBT_MAGIC_BYTES = Buffer.from([0x70, 0x73, 0x62, 0x74, 0xff]);
```
bucko13, line 85: might be worth just adding this with the existing PSBT_MAGIC consts in the `psbt.js` file?

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+    for (const map of this.outputMaps) {
+      serializeMap(map, bw);
+    }
+
+    return bw.render().toString(format);
+  }
+
+
+  // Attempt to return a PsbtV2 by converting from a PsbtV0 string or Buffer
+  static FromV0 (psbt:string|Buffer):PsbtV2 {
```
bucko13, line 853: I guess this is capitalized because it's `static`? Other static methods in the library are still camel cased. I've not seen pascal casing for anything but objects/classes personally.

src/psbtv2.ts
bucko13: I might be missing something obvious, but why is it that these need to be done in two steps (same for the outputs)? What's the difference between what comes back from `data.inputs` and `txInputs`? Again, probably missing something, but feels like it would be nice to do it in one step.

Separate issue and might be an over optimization, but might be nice to have these as standalone `addInputsFromPsbtV0` and `addOutputsFromPsbtV0` not a huge deal though.

bucko13 (changes requested):
Mostly nits and suggestions for consolidation with existing tools. Looking good so far though!

src/psbtv2.ts
Shadouts: It's not just that. Unless you have save on change/pause turned on, you get a bunch of red lines until it does get saved. Manually saving just to check if I really did make a syntax mistake is a bit annoying. I was planning to add a prettier config to address tyler's request for type annotation spacing.

src/psbtv2.ts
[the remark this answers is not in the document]
tylerchambers: Yeah I might have hosed the regex, main concern is changing to an early return  and cutting out multiple if typeof psbt === "string" checks if possible for the sake of readability.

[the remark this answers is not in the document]
Shadouts: Done. I usually try to avoid a multiple return kind of flow, but this function is short enough.

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+
+    // Build globalMap
+    while(readAndSetKeyPair(this.globalMap, br));
+    if (
+      this.PSBT_GLOBAL_TX_VERSION === undefined || 
+      this.PSBT_GLOBAL_INPUT_COUNT === undefined || 
+      this.PSBT_GLOBAL_OUTPUT_COUNT === undefined ||
+      this.globalMap.has('00') // PsbtV2 must exclude key 0x00
+    ) {
+      throw Error("Provided psbtV2 not valid. Missing required global values.");
```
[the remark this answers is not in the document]
Shadouts: I think that only helps when catching an `Error` and rethrowing, otherwise they are functionally the same. Are you missing stake traces on these?

src/psbtv2.ts
```diff
@@ -0,0 +1,910 @@
...
+// A getter helper for optional keytypes which returns lists of values as hex
+// strings.
+function getOptionalMappedBytesAsHex (maps:Map<Key, Value>[], keytype:KeyType) {
+  const bytes:(string|null)[] = [];
+
+  for (const map of maps) {
+    bytes.push(map.get(keytype)?.toString('hex') ?? null);
+  }
+
+  return bytes;
```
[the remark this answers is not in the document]
Shadouts: Bordering on being too one-liner. I'm having flashbacks from the jquery days, but I'll add it.

src/psbtv2.ts
Shadouts: This is reading from a `bitcoinjs-lib` Psbt. Apparently not all values are available from one attribute.

src/psbtv2.ts
bucko13: Take a look at this PR: https://github.com/unchained-capital/unchained-bitcoin/pull/60

hopefully we can get that in and automate a lot of this. Once it’s merged to master I’ll rebase back to the feature branches.

src/psbtv2.ts
Shadouts: @tylerchambers, similarly, this `entries` function is returning something that is not an array prototype. It has no "map".

src/psbtv2.ts
bucko13: I also tested that with that the spacing is taken care of.

src/psbtv2.ts
Shadouts: See above

src/psbtv2.ts
Shadouts: added

src/psbtv2.ts
Shadouts: done

src/psbtv2.ts
Shadouts: The `validateBIP32Path` is useful. The paths module doesn't export the regex values and they aren't as useful as what is here. The regex I use here lists nodes as match groups in a single operation. Additionally, the hardening offset is a constant instead of requiring an operation from `Math`.

I'd argue that it's probably the paths module which needs to be updated/improved with these features, but should it fit in the scope of the ledger-v2 feature?

bucko13: Agree yours is probably better.

As for scope, it's a tough question. You don't want to introduce scope creep but if you need an improvement to an existing feature to improve the feature within scope, and the scope creep isn't out of control, I would say it's probably worth it. We likely wouldn't have other opportunities to explicitly be fixing/updating something like this except within the scope of other issues. Similar also to increasing typescript coverage for example.

Shadouts:
I just pushed some changes to the constructor which adds validation and an `nLockTime` getter both required to pass BIP0370 tests.

Shadouts:
I'm suspicious about `Psbt.nLockTime`. I feel like something is off about it, but the algorithm description in BIP0370 isn't entirely clear. I attempted to follow what I thought it was saying, but it was failing tests, so I adjusted it. To be specific, I thought it was saying to only respect the locktime type that is present in ALL inputs, but it seems to require favoring a majority.

src/psbtv2.ts
```diff
@@ -676,6 +690,63 @@ export class PsbtV2 {
...
+        heights.push(heightLocks[i] as number);
+      } 
+      
+      if (timeLocks[i] !== null) {
+        times.push(timeLocks[i] as number);
+      }
+    }
+
+    // From BIP0370: If none of the inputs have a PSBT_IN_REQUIRED_TIME_LOCKTIME
+    // and  PSBT_IN_REQUIRED_HEIGHT_LOCKTIME, then PSBT_GLOBAL_FALLBACK_LOCKTIME
```
bucko13, line 699: should this be "or"? so if none of the inputs have time OR height locktime?

src/psbtv2.ts
```diff
@@ -676,6 +690,63 @@ export class PsbtV2 {
...
+    }
+
+    // From BIP0370: If one or more inputs have a PSBT_IN_REQUIRED_TIME_LOCKTIME
+    // or PSBT_IN_REQUIRED_HEIGHT_LOCKTIME, then the field chosen is the one
+    // which is supported by all of the inputs. This can be determined by
+    // looking at all of the inputs which specify a locktime in either of those
+    // fields, and choosing the field which is present in all of those inputs.
+    // Inputs not specifying a lock time field can take both types of lock
+    // times, as can those that specify both. The lock time chosen is then the
+    // maximum value of the chosen type of lock time.
```
bucko13, line 713: This is what you're saying you weren't sure about in your PR comment? I think the thinking goes that only the largest one will hold any "authority". Basically until that lock time is passed then it doesn't matter what the others say. I know there is some bad edge case where you can have an input that locks by block and one that locks by time and that they can be in conflict, but i don't think any of that is of concern here.

bucko13: actually might just be a complication within a single script so slightly different: https://medium.com/blockstream/dont-mix-your-timelocks-d9939b665094

src/psbtv2.ts
Shadouts: Probably, but I just copied it from https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki#determining-lock-time

bucko13 (approved):
This looks good to me! just a couple small nits and questions but I think this is good to go.

src/psbtv2.ts
Shadouts: Fixed

bucko13: ha, fair enough

src/psbtv2.ts
Shadouts: That isn't the line I was concerned with. Take this excerpt from the bip:
```
If one or more inputs have a PSBT_IN_REQUIRED_TIME_LOCKTIME
or PSBT_IN_REQUIRED_HEIGHT_LOCKTIME, then the field chosen is the one
which is supported by all of the inputs. This can be determined by
looking at all of the inputs which specify a locktime in either of those
fields, and choosing the field which is present in all of those inputs.
Inputs not specifying a lock time field can take both types of lock
times, as can those that specify both. The lock time chosen is then the
maximum value of the chosen type of lock time.

If a PSBT has both types of locktimes possible because one or more inputs
specify both PSBT_IN_REQUIRED_TIME_LOCKTIME and
PSBT_IN_REQUIRED_HEIGHT_LOCKTIME, then locktime determined by looking at
the PSBT_IN_REQUIRED_HEIGHT_LOCKTIME fields of the inputs must be chosen.
```

I tried only applying a locktime other than global fallback when all inputs have a locktime specified (as the bip seems to describe when it says ` If one or more inputs have a PSBT_IN_REQUIRED_TIME_LOCKTIME
or PSBT_IN_REQUIRED_HEIGHT_LOCKTIME, then the field chosen is the one
which is supported by all of the inputs`), but one/some of the tests were failing. Based on the test vectors, perhaps what the the bip is trying to describe is to look at all inputs which have a locktime specified, then from those pick the one which is supported by all those inputs? I think I tried that and tests were still failing, but maybe I didn't do it correctly.

The way I've done it is to check and see if heights is supported by all and if so use that, otherwise check for the majority type.

src/psbtv2.test.js
```diff
@@ -0,0 +1,857 @@
...
+    base64:
+      "cHNidP8BAgQCAAAAAQQBAQEFAQIBBgECAfsEAgAAAAABAFICAAAAAcGqJW4hS5ahgi+T3kK/87Xz/40FGTBuNRXXUVpegFsSAAAAAAD/////ARjGmjsAAAAAFgAUsKOvFEIIQSaTyn0WaFK1LbCu8G4AAAAAAQEfGMaaOwAAAAAWABSwo68UQghBJpPKfRZoUrUtsK7wbgEOIAsK2SFBnByHGXNdctxzn56p4GONH+TB7vD5lECEgV/IAQ8EAAAAAAAiAgLWAfhIRqZ1X3dr4A49nej7EKzJNfuDxF+wFi1MrVq3khj2nYc+VAAAgAEAAIAAAACAAAAAACoAAAABAwgACK8vAAAAAAEEFgAUxDD2TEdW2jENvRoIVXLvKZkmJywAIgIC42+/9T3VNAcM+P05ZhRoDzV6m4Xbc0C/HPp0XSrXs0AY9p2HPlQAAIABAACAAAAAgAEAAABkAAAAAQMIi73rCwAAAAABBBYAFE3Rk6yWSlasG54cyoRU/i9HT4UTAA==",
+    inputs: 1,
+    outputs: 2,
+    modifiable: ["OUTPUTS"],
+  },
+
+  // Case: 1 input, 2 output updated PSBTv2, with Has SIGHASH_SINGLE Flag (bit
```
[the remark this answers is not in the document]
Shadouts: Copied directly from BIP0370

src/psbtv2.ts
```diff
@@ -0,0 +1,953 @@
...
+function getOptionalMappedBytesAsUInt(
+  maps: Map<Key, Value>[],
+  keytype: KeyType
+) {
+  return maps.map((map) => map.get(keytype)?.readUInt32LE() ?? null);
+}
+
+// Accepts a BIP0032 path as a string and returns a Buffer containing uint32
+// values for each path node.
+function parseDerivationPathNodesToBytes(path: string): Buffer {
```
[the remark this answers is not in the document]
Shadouts: It's just not as good. I'd rather replace it since I already had built one which, imo, does a more complete job, but I was thinking that increases scope too much.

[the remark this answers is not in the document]
bucko13: right. got it!

src/psbtv2.test.js
```diff
@@ -0,0 +1,857 @@
...
+    base64:
+      "cHNidP8BAgQCAAAAAQQBAQEFAQIB+wQCAAAAAAEAUgIAAAABwaolbiFLlqGCL5PeQr/ztfP/jQUZMG41FddRWl6AWxIAAAAAAP////8BGMaaOwAAAAAWABSwo68UQghBJpPKfRZoUrUtsK7wbgAAAAABAR8Yxpo7AAAAABYAFLCjrxRCCEEmk8p9FmhStS2wrvBuAQ4gCwrZIUGcHIcZc11y3HOfnqngY40f5MHu8PmUQISBX8gBDwQAAAAAARIEAGXNHQAiAgLWAfhIRqZ1X3dr4A49nej7EKzJNfuDxF+wFi1MrVq3khj2nYc+VAAAgAEAAIAAAACAAAAAACoAAAABAwgACK8vAAAAAAEEFgAUxDD2TEdW2jENvRoIVXLvKZkmJywAIgIC42+/9T3VNAcM+P05ZhRoDzV6m4Xbc0C/HPp0XSrXs0AY9p2HPlQAAIABAACAAAAAgAEAAABkAAAAAQMIi73rCwAAAAABBBYAFE3Rk6yWSlasG54cyoRU/i9HT4UTAA==",
+  },
+];
+
+const BIP_370_VECTORS_VALID_PSBT = [
+  // Index 0
```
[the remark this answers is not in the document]
Shadouts: Literally the index of the array. Further down there is an `Index 14` where the timelock tests start.

src/psbtv2.ts
bucko13: > but one/some of the tests were failing. Based on the test vectors, perhaps what the the bip is trying to describe is to look at all inputs which have a locktime specified, then from those pick the one which is supported by all those inputs?

Yeah that's how I read it, but agree it's very confusing.

src/psbtv2.test.js
```diff
@@ -0,0 +1,857 @@
+import { PsbtV2 } from "./psbtv2";
+
+const BIP_370_VECTORS_INVALID_PSBT = [
+  // Case: PSBTv0 but with PSBT_GLOBAL_VERSION set to 2.
```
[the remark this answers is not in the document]
Shadouts: Maybe I'll try to figure that out.

src/psbtv2.test.js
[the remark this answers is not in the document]
bucko13: Yeah, I saw it later on how you're pulling in explicit indexes, so this is just a way to keep track.

bucko13 (approved):
Nice!

PR #61, opened 2023-02-17
feat: adds json util methods to braid

bucko13:
this is just a small feature update that would've been nice for some of the tooling I built out for wallet registration in https://github.com/unchained-capital/unchained-wallets/pull/89

Most of the changes in this file are from prettier. the `toJSON` and `fromJSON` are the only relevant/functional changes.


humanumbrella (approved):
lgtm

PR #62, opened 2023-02-18
Psbt v2 maps

Shadouts:
If other classes might need restricted PsbtV2 style serialization and mapping but must remain functionally separate from a controlled PsbtV2 flow, they can utilize the abstract class PsbtV2Maps.


bucko13:
I was able to get this working to sign with the ledger v2 app using unchained-wallets with two changes.

First is it needs a `copy` method. Copying the implementation from ledger-bitcoin worked:

```typescript
class PsbtV2 {
  ...
  public copy(to: PsbtV2) {
    this.copyMap(this.globalMap, to.globalMap);
    this.copyMaps(this.inputMaps, to.inputMaps);
    this.copyMaps(this.outputMaps, to.outputMaps);
  }

  private copyMaps(
    from: readonly ReadonlyMap<string, Buffer>[],
    to: Map<string, Buffer>[]
  ) {
    from.forEach((m, index) => {
      const to_index = new Map();
      this.copyMap(m, to_index);
      to[index] = to_index;
    });
  }

  // eslint-disable-next-line class-methods-use-this
  private copyMap(from: ReadonlyMap<string, Buffer>, to: Map<string, Buffer>) {
    from.forEach((v, k) => to.set(k, Buffer.from(v)));
  }
}
```

Second is that I ran into a Buffer offset error when trying to convert from a PSBTv0 to a v2. I was able to fix it by changing line 904 in the following way (I imagine there's probably a better way to resolve this, but it worked... 🤷‍♂️ ):

```javascript
// psbtv2.PSBT_GLOBAL_TX_VERSION = psbtv0.data.getTransaction().readInt32LE(); <--- error
psbtv2.PSBT_GLOBAL_TX_VERSION = psbtv0.data.globalMap.unsignedTx.tx.version; // <--- worked
```

PR #65, opened 2023-02-23
Wallet config updates

bucko13:
adds a new tool for getting a masked derivation that's used in places like unchained-wallets for configurations.

### DIFF
--- a/src/keys.test.js
+++ b/src/keys.test.js
@@ -16,7 +16,8 @@ import {
   ExtendedPublicKey,
   fingerprintToFixedLengthHex,
   extendedPublicKeyRootFingerprint,
-  validateExtendedPublicKeyForNetwork
+  validateExtendedPublicKeyForNetwork,
+  getMaskedDerivation,
 } from "./keys";
 
 import { TESTNET, MAINNET } from "./networks";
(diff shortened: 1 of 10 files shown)


src/keys.test.js
```diff
@@ -16,7 +16,8 @@ import {
   ExtendedPublicKey,
   fingerprintToFixedLengthHex,
   extendedPublicKeyRootFingerprint,
-  validateExtendedPublicKeyForNetwork
+  validateExtendedPublicKeyForNetwork,
```
nk1tz, line 19: Are we not properly setup yet for jest+typescript?

bucko13: I don't think so. I forget what the issue was as this was setup a while ago, but only a couple of files are typescript right now anyway.

src/types/addresses.ts
```diff
@@ -0,0 +1,5 @@
+// P2TR not able to be used anywhere yet but technicall a valid multisig
```
nk1tz, line 1: nip typo`technically`

nk1tz (approved):

Shadouts (approved):

PR #66, opened 2023-02-23
feat: better PsbtV2 validation to ensure proper PsbtV2s

Shadouts:
Somewhat of a prerequisite for an `addSignature` method. Different roles have different permissions. Strictly controlling these will ensure a psbt in a valid state. This change refactors and adds some validation logic for the Creator and Constructor roles.


bucko13 (approved):
very nice.

PR #69, opened 2023-02-28
Fix PsbtV2.addInput, set PSBT_GLOBAL_TX_VERSION, and adjust some variable names

Shadouts:
*This is a minor version increment as it adds a breaking change on a beta lib to `PsbtV2.addInput` arguments. It also changes the function of `PsbtV2.FromV0` resulting in a PsbtV2 instance different from what it previously would have been due to the changed handling of set PSBT_GLOBAL_TX_VERSION*

This PR addresses three things: 

First, `PsbtV2.addInput()` was fixed so that it properly handles the setting of `PSBT_IN_WITNESS_UTXO`.  Previously, it was only setting this value as the witness script. Now it should be setting it as `<64-bit little endian int amount> <compact size uint scriptPubKeylen> <bytes scriptPubKey>` per BIP0174.

Second, the PSBT_GLOBAL_TX_VERSION setter was adjusted to default to 2 if a value less than 2 is passed in. Upon creation, a PsbtV2 must have a PSBT_GLOBAL_TX_VERSION gte 2. This mostly should affect `PsbtV2.FromV0` which may be passed a psbt with this value set to 1.

Third, a few badly named variables were fixed.

While this change fixes a Ledger error in Caravan's test suite to do with signing of native segwit txns, it creates a failure in the signing tests (including non-segwit) due to the txn version being updated. The text fixtures there might need to be adjusted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "unchained-bitcoin",
-  "version": "0.2.1",
+  "version": "0.3.0",
   "description": "Unchained Capital's bitcoin utilities library.",
   "main": "lib/index.js",
   "repository": {
--- a/src/psbtv2.ts
+++ b/src/psbtv2.ts
@@ -358,8 +358,14 @@ export class PsbtV2 extends PsbtV2Maps {
   }
 
   set PSBT_GLOBAL_TX_VERSION(version: number) {
+    let workingVersion = version;
+    if (workingVersion < 2) {
+      console.warn(`PsbtV2 cannot have a global tx version less than 2. Version ${workingVersion} specified. Setting to version 2.`);
+      workingVersion = 2;
+    }
+
     const bw = new BufferWriter();
-    bw.writeI32(version);
+    bw.writeI32(workingVersion);
     this.globalMap.set(KeyType.PSBT_GLOBAL_TX_VERSION, bw.render());
   }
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "unchained-bitcoin",
-  "version": "0.2.1",
+  "version": "0.3.0",
```
bucko13, line 3: Thoughts on leaving this for a separate commit on `master`?

bucko13 (commented):
want to do some testing with the version enforcement. this is definitely something we want to have sooner rather than later though.

src/psbtv2.ts
```diff
@@ -358,8 +358,14 @@ export class PsbtV2 extends PsbtV2Maps {
   }
 
   set PSBT_GLOBAL_TX_VERSION(version: number) {
+    let workingVersion = version;
+    if (workingVersion < 2) {
+      console.warn(`PsbtV2 cannot have a global tx version less than 2. Version ${workingVersion} specified. Setting to version 2.`);
+      workingVersion = 2;
+    }
+
```
bucko13, line 366: I wanted to note here that I think we should be careful how we handle this. 

Just for background, here's the language from the [BIP](https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki#user-content-Constructor): 

It [says](https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki#cite_ref-3-0):
> The transaction version number must be set to at least 2.

To add more clarity, the reason for this is given:

> Why does the transaction version number need to be at least 2? The transaction version number is part of the validation rules for some features such as OP_CHECKSEQUENCEVERIFY. Since it is backwards compatible, and there are other ways to disable those features (e.g. through sequence numbers), it is easier to require transactions be able to support these features than to try to negotiate the transaction version number.

So the reason is essentially to make negotiation easier for special features. 

With this context, I think that changing the version in the setter is probably not the right approach. At the very least, I would say that if we think this is a hard rule that should be stuck to, then we should throw an error here rather than maleate the transaction essentially (this is particularly problematic if you're converting from V0). 

On the other hand, it seems like other implementations are not very opinionated on this and it also does break backwards compatibility in a sense which was another stated goal of BIP370. 

All that to say, I think I would argue for some way to at least allow for versions less than 2, at least in some narrow edge cases. Maybe we can accept an argument `allowLegacy: boolean` that allows a version to be set to < 2. If this is not `true` then the method will throw, and if it is true then a warning is logged but the version is set to the "wrong" version.

bucko13 (changes requested):

Shadouts:
Closing this. I'll be breaking this PR apart into smaller pieces.

PR #71, opened 2023-02-28
Fix confusing variable names in PsbtV2

Shadouts:
Some variable names were improper and/or copied from other method definitions. This changes them to be less confusing.


bucko13 (approved):
Send it!

PR #72, opened 2023-02-28
PsbtV2.PSBT_GLOBAL_TX_VERSION validation and txnv1 compatibility util

Shadouts:
For reasons, a PSBTv2 is invalid if the PSBT_GLOBAL_TX_VERSION is less than 2 (https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki#cite_note-3). This change enforces such a requirement by throwing if a Creator role (or any other role) attempts to set the value less than 2.

Other implementations may ignore this specification by allowing a PSBTv2 with PSBT_GLOBAL_TX_VERSION set to 1. Compatibility with these implementations may be desired while working with an already created and partially signed psbtV0 on txnv1. To provide this, the static `PsbtV2.FromV0` now takes a bool argument `allowTxnVersion1` which will allow for PSBT_GLOBAL_TX_VERSION to be dangerously set to 1.

### DIFF
--- a/src/psbtv2.test.js
+++ b/src/psbtv2.test.js
@@ -913,19 +913,27 @@ describe("PsbtV2.FromV0", () => {
     (vect) => {
       for (const key in vect) {
         if (key === "hex" || key === "base64") {
-          const t = () => PsbtV2.FromV0(vect[key]);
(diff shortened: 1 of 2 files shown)


bucko13 (approved):
Looks good! Thanks.

PR #73, opened 2023-02-28
Upgrade bufio to 1.2.0 for browser reads on PsbtV2.PSBT_OUT_AMOUNT

Shadouts:
The bufio package supports `BufferReader.readBigI64` since version 1.2.0. This change upgrades the package and fixes the `PsbtV2.PSBT_OUT_AMOUNT` getter so that browsers can use it.


bucko13 (approved):
Tested this with unchained-wallets and in caravan test suite and can confirm this works.

PR #74, opened 2023-03-02
fix: get input data based on address type

bucko13:
Some eslint fixes got caught in this but basically the input data for adding to a psbt needs to be a bit smarter about how it gets the input data for ledger v2 which is pickier about when it requires nonWitnessUtxo vs witnessUtxo data. (See [here](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#simple-signer-algorithm) from the bip174 docs).

Unfortunately while this creates a valid PSBT that ledger will sign, the signatures created after getting converted to psbtv2 and getting passed to the ledger are still wrong so there's still some tweaking left to do. 

Can test with: https://github.com/unchained-capital/caravan/pull/294

### DIFF
--- a/src/psbt.js
+++ b/src/psbt.js
@@ -1,20 +1,20 @@
-import {Psbt, Transaction} from "bitcoinjs-lib";
-import { reverseBuffer } from 'bitcoinjs-lib/src/bufferutils';
-import {toHexString} from './utils';
+import { Psbt, Transaction } from "bitcoinjs-lib";
+import { reverseBuffer } from "bitcoinjs-lib/src/bufferutils";
+import { toHexString } from "./utils";
 import {
   generateMultisigFromHex,
   multisigAddressType,
   multisigBraidDetails,
   multisigRedeemScript,
   multisigWitnessScript,
-} from './multisig';
-import {bip32PathToSequence} from './paths';
-import BigNumber from 'bignumber.js';
-import {P2SH} from './p2sh';
-import {P2WSH} from './p2wsh';
-import {P2SH_P2WSH} from './p2sh_p2wsh';
-import {generateBip32DerivationByIndex, generateBraid} from './braid';
-import {networkData} from './networks';
+} from "./multisig";
+import { bip32PathToSequence } from "./paths";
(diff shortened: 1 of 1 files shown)


src/psbt.js
```diff
@@ -200,13 +202,20 @@ export function psbtInputFormatter(input) {
...
-  const utxoToVerify = isSegWit ? {witnessUtxo} : {nonWitnessUtxo};
+  const addressType = multisigAddressType(input.multisig);
+
+  // The PSBT class in bitcoinjs-lib won't accept both nonWitnessUxo and witnessUtxo
+  // but ledger v2 needs nonWitnessUtxo data in the PSBT for any transactions that aren't
+  // native segwit. So we need to pick which one to give based on the address type
+  // https://github.com/bitcoinjs/bitcoinjs-lib/issues/1595
+  const utxoToVerify =
+    addressType === P2WSH ? { witnessUtxo } : { nonWitnessUtxo };
+
```
bucko13, line 218: this is basically the only relevant change

Shadouts (approved):

Shadouts (changes requested):
I know we talked about this out-of-band, but there are still some signature test failures on this which is to be expected since the psbt contents are changing.

Shadouts:
I went to go see about the sigs myself, but I'm still getting invalid data error `0x6a80`

bucko13:
@Shadouts thats weird…

bucko13:
So, the correct behavior for this is that if it's `P2SH-P2WSH` then BOTH nonWitnessUtxo and witnessUtxo should be included in the PSBT. Unfortunately bitcoinjs-lib doesn't allow you to do this though. I've filed a [bug](https://github.com/bitcoinjs/bitcoinjs-lib/issues/1894). Hopefully it'll be addressed and we can fix the input serialization in the correct way such that signatures are created correctly. Without this change, ledger will fail b/c it has incorrect data. With this change, ledger will succeed but will produce incorrect signatures so it's not quite ready for primetime.

bucko13:
Not required anymore for now. Closing.

PR #75, opened 2023-03-06
PsbtV2.addPartialSig()

Shadouts:
This change fixes a few bugs as well as provides the `PsbtV2.addPartialSig()` method. Per BIP0174, the Signer is supposed to add any signature it creates to the psbt, however, some signing devices such as the Ledger may only return the signature instead of the full psbt with signatures added. This method provides a way to add the signature to the psbt to aid the Signer role.  This method does not perform any cryptographic validation to check that the values satisfy the input. It accepts any value and sets them to the input map at the specified index.

This method calls a `private PsbtV2.handleSighashType()` which performs necessary updates to `PSBT_GLOBAL_TX_MODIFIABLE` as defined at https://github.com/bitcoin/bips/blob/master/bip-0370.mediawiki#signer

### DIFF
--- a/src/psbtv2.test.js
+++ b/src/psbtv2.test.js
@@ -921,10 +921,11 @@ describe("PsbtV2.FromV0", () => {
   );
 
   it("Throws with valid psbtv0 with PSBT_GLOBAL_TX_VERSION=1 when disallowing txnv1", () => {
-    const t = () => PsbtV2.FromV0(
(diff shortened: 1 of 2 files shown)


src/psbtv2.ts
```diff
@@ -1026,19 +1057,93 @@ export class PsbtV2 extends PsbtV2Maps {
...
+    const br = new BufferReader(sig.slice(-1));
+    let sighashVal = br.readU8();
+    let modifiable = this.PSBT_GLOBAL_TX_MODIFIABLE;
+
+    if (!(sighashVal & SighashType.SIGHASH_ANYONECANPAY)) {
+      modifiable = modifiable.filter(
+        (val) => val !== PsbtGlobalTxModifiableBits.INPUTS
+      );
+    } else {
+      // Remove SIGHASH_ANYONECANPAY bit for simpler comparisons
```
bucko13, line 1111: can you explain this comment? I don't really follow.

src/psbtv2.ts
```diff
@@ -1026,19 +1057,93 @@ export class PsbtV2 extends PsbtV2Maps {
...
+    if (sighashVal !== SighashType.SIGHASH_NONE) {
+      modifiable = modifiable.filter(
+        (val) => val !== PsbtGlobalTxModifiableBits.OUTPUTS
+      );
+    }
+    if (
+      sighashVal === SighashType.SIGHASH_SINGLE &&
+      !modifiable.includes(PsbtGlobalTxModifiableBits.SIGHASH_SINGLE)
+    ) {
+      console.log("inside");
```
bucko13, line 1125: nit: cleanup

src/psbtv2.test.js
```diff
@@ -988,3 +989,107 @@ describe("getPsbtVersionNumber", () => {
...
+    expect(psbt.PSBT_IN_PARTIAL_SIG[0].length).toBe(0);
+  });
+
+  it("Can add successfully", () => {
+    psbt.addInput({ previousTxId: Buffer.from([0x00]), outputIndex: 0 });
+    psbt.addInput({ previousTxId: Buffer.from([0x01]), outputIndex: 1 });
+    psbt.addPartialSig(0, Buffer.from([0x00]), Buffer.from([0x00]));
+    psbt.addPartialSig(0, Buffer.from([0x01]), Buffer.from([0x01]));
+    psbt.addPartialSig(1, Buffer.from([0x02]), Buffer.from([0x02]));
+    console.log(psbt.PSBT_IN_PARTIAL_SIG);
```
bucko13, line 1036: nit: cleanup

bucko13 (changes requested):
Looks good! Just a couple of small cleanup nits but otherwise should be good (as far as I can tell at least)

src/psbtv2.ts
Shadouts: The sighash flag seems to combine and enumeration with a bitmask. If bit 8 is set, it's `SIGHASH_ANYONECANPAY`. The other three flags are enumerated on values of 1, 2, and 3. It was easier to just set bit 8 to 0 and then compare by `value === 3` then to instead have to compare by `value === 3 || value === 131`.

bucko13: ok, gotcha.

bucko13 (approved):

PR #76, opened 2023-03-07
Set partialSigs from PsbtV2.FromV0

Shadouts:
Sometimes when converting from a psbtV0, valid partialsigs might already be present. This change would add them to the resultant psbtV2.

### DIFF
--- a/src/psbtv2.ts
+++ b/src/psbtv2.ts
@@ -1,5 +1,3 @@
-/* eslint-disable no-shadow */
-
 /**
  * The PsbtV2 class is intended to represent an easily modifiable and
  * serializable psbt of version 2 conforming to BIP0174. Getters exist for all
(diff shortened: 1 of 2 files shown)


src/psbtv2.ts
```diff
@@ -982,8 +983,7 @@ export class PsbtV2 extends PsbtV2Maps {
       }
     }
 
-    this.inputMaps.push(map);
-    this.PSBT_GLOBAL_INPUT_COUNT = this.inputMaps.length;
+    this.PSBT_GLOBAL_INPUT_COUNT = this.inputMaps.push(map);
```
bucko13, line 986: clever! the fact that `push` returns the length usually is the cause of bugs for me because I forget about it. Nice to see it used deliberately.

bucko13 (approved):
Awesome. good stuff

PR #77, opened 2023-03-08
PsbtV2 additional PSBT_GLOBAL_TX_MODIFIABLE validation, PsbtV2.removePartialSig, and PsbtV2.isModifiable

Shadouts:
This change seeks to provide more control on when inputs and outputs are allowed to be modified based on `PSBT_GLOBAL_TX_MODIFIABLE`. 

Part of adding this validation required adding a `removePartialSig` method. When a psbt is set for `SIGHASH_SINGLE`, removing an output will invalidate a sig, so the sig must be removed. Perhaps this might need to be constrained further if it causes problems.

The `FromV0` static method also needed to be adjusted to add the sigs at the very end. Adding a signature has the potential to unset the ability to modify inputs or outputs, so they are added last after the Constructor role has added all the previously existing inputs and outputs.

### DIFF
--- a/src/psbtv2.test.js
+++ b/src/psbtv2.test.js
@@ -1015,6 +1015,9 @@ describe("PsbtV2.addPartialSig", () => {
   beforeEach(() => {
     psbt = new PsbtV2();
     psbt.handleSighashType = jest.fn();
+    // This has to be added so inputs can be added else addSig will fail since
+    // the input at the index does not exist.
+    psbt.PSBT_GLOBAL_TX_MODIFIABLE = ["INPUTS"];
   });
 
   it("Throws on validation failures", () => {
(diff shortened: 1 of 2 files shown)


src/psbtv2.test.js
```diff
@@ -1052,14 +1055,73 @@ describe("PsbtV2.addPartialSig", () => {
...
+    psbt = new PsbtV2();
+    psbt.handleSighashType = jest.fn();
+    // This has to be added so inputs can be added else removeSig will fail
+    // since the input at the index does not exist.
+    psbt.PSBT_GLOBAL_TX_MODIFIABLE = ["INPUTS"];
+  });
+
+  it("Throws on validation failures", () => {
+    const removeSig = (index, pub, sig) =>
+      psbt.removePartialSig(index, pub, sig);
```
bucko13, line 1081: I don't think `sig` is actually passed in either test here. so could just be 
```suggestion
    const removeSig = (index, pub) =>
      psbt.removePartialSig(index, pub);
```

bucko13 (approved):
LGTM! 🚀

PR #78, opened 2023-03-13
fix(fixture): missing fixture for testnet xpub export

bucko13:
Adding this gets uc-bitcoin back in line with Caravan test suite to cover a missing fixture that were causing some tests to fail. 
Caravan will need its dependency bumped. 

This test is currently failing:
Screenshot 2023-03-13 at 10 22 21 AM


humanumbrella (approved):

waldenraines (approved):
LGTM

PR #81, opened 2023-03-31
automate release process using github actions

bucko13:


.github/.workflows/release-please.yml
```diff
@@ -0,0 +1,19 @@
...
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        with:
+          release-type: node
+          package-name: release-please-action
```
sashafklein, line 19: I _think_ `package-name` should be `unchained-bitcoin`. I also think you need `command: manifest` here, so that the action looks at the `release-please-config.json` for deets.

sashafklein (changes requested):

.github/.workflows/release-please.yml
bucko13: amended with those changes. thanks

sashafklein (approved):

PR #85, opened 2023-04-06
Add minSats to the validateOutputAmount function. Fix missing package.

Shadouts:
Adjusts `validateOutputAmount` to accept an optional minSats which allows a caller to specify their own dust limit.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@babel/polyfill": "^7.7.0",
         "bignumber.js": "^8.1.1",
         "bip32": "^2.0.5",
+        "bip66": "^1.1.5",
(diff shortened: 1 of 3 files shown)


bucko13 (changes requested):
I might be missing it in the formatting changes in the diff but I don't think the new functionality has tests.

Shadouts:
> I might be missing it in the formatting changes in the diff but I don't think the new functionality has tests.

https://github.com/unchained-capital/unchained-bitcoin/pull/85/files#diff-39ce24c877d13219407c3baecd4639a5b9234bcc55437f1608abf4d006d03cbaR94-R96

but I see it's tested again below there. I will add another one.

bucko13 (approved):
Looks good!

PR #87, opened 2023-05-30
fix(braid): allow for regtest braids

bucko13:


src/addresses.test.js
```diff
@@ -1,66 +1,75 @@
...
+            invalidAddress
+          );
         });
       });
     });
 
-    it("returns an empty string when the address is valid",  () => {
-      Object.values(NETWORKS).forEach((network) => {
+    it("returns an empty string when the address is valid", () => {
+      [NETWORKS.MAINNET, NETWORKS.TESTNET].forEach((network) => {
```
bucko13, line 66: This was the only real change in this file.

tylerchambers (approved):

PR #89, opened 2023-08-04
feat: ensure BigNumber classes are not returned

Shadouts:
This would be a breaking change.

A few functions in this library were returning `BigNumber` objects. In some cases, these functions were mistyped by jsdoc to report that they should be returning a number. This creates an API dependency requiring any consumers of the library to be concerned with aspects of `BigNumber`. This change ensures that `BigNumbers` are converted to a string representation before returning. Consumers of the library can then rely on their preferred big number library (such as `Big.js`).

This change also takes the position that big numbers should be passed as string representations even if a number representation is possible in order to deter arithmetic operations without first employing a big number handling library of choice.

Key functions which are affected by this change:
- `estimateMultisigTransactionFeeRate`
- `estimateMultisigTransactionFee`
- `bitcoinsToSatoshis`
- `satoshisToBitcoins`

Other unchained-capital libraries likely needing an update due to this change:
- https://github.com/unchained-capital/caravan
- https://github.com/unchained-capital/unchained-wallets


src/fees.test.js
bucko13, line 1: would it be a big change to convert this to ts as well?

src/fees.ts
```diff
@@ -1,35 +1,33 @@
...
-  P2WSH,
-  estimateMultisigP2WSHTransactionVSize,
-} from "./p2wsh";
-import {ZERO} from "./utils";
+import { P2WSH, estimateMultisigP2WSHTransactionVSize } from "./p2wsh";
+import { ZERO } from "./utils";
+
+// Without this, BigNumber will report strings as exponentials. 16 places covers
+// all possible values in satoshis.
+BigNumber.config({ EXPONENTIAL_AT: 16 });
```
bucko13, line 20: Is there any concern about this impacting uses of BigNumber that aren't for satoshis?

bucko13 (changes requested):
I think this is a great change fixing a pretty big oversight. I think continued conversion to typescript across the OSS libraries will help continue pushing us in these cleaner directions as well. 

Couple comments in addition to the test.js-> .ts comment:
- I think the commit should have a prefix `feat!` so that the automatic release does a major release since this is a breaking change. 
- Could you add to the PR description the other OSS libraries in `unchained-capital/` that will need updates to support the change?

src/fees.test.js
Shadouts: This would require a more significant change perhaps better suited for another PR. 

Short answer: There is currently no support for typescript tests in this repo.

Long answer: The repo is currently testing the transpiled javascript files. The transpiler transpiles and moves them into the `lib` folder and then also moves the test files into the `lib` distributable, then `jest` is run against them. To me, this seems a bit strange. Essentially, the tests are testing not only the logic contained in the source, but the transpiler. Perhaps for software that controls money this is important to be sure that the transpiler didn't break something, but the transpiler is transpiling the test files too! So, I don't think anything is gained by converting the test files to typescript.

If the distributable should be tested, I think JS test files are sufficient. If the logic should be tested, then typescript test support should be installed and package.json should be updated to test before transpiling.

bucko13: I don’t understand what we lose by making this a ts file? What we gain is that at least we have type safety on the functions being called/tested or am I misunderstanding something? Even if we’re testing js files, going through the transpiler alone gives some marginal benefit.

I’m not disagreeing with the need for your longer term proposal, but I don’t see what we lose by capturing test files in the incremental conversion in the meantime. If nothing else it also makes the truly needed change your proposing slightly smaller in scope since some files will already be converted. 

If I’m missing some additional context about why this would be more costly than beneficial in the meantime though I’m happy to be convinced otherwise.

bucko13: IDE benefits as well which are nice post-conversion

Shadouts: First: To simply change this to typescript requires that more libraries must be installed increasing the scope of the proposed change. My environment is yelling at me about not having types installed for jest. Let's do it in another PR.

As for the philosophy of it, I see testing the distributable similar to a contract test. The test should be emulating how a consumer of the library might be running tests against the library's API. There are no guarantees that the consumer of a library would be using typescript. In this sense, what testing library is used shouldn't matter. There's nothing lost by switching to typescript tests. I'm arguing that the gains are superficial.

I favor unit testing the pre-distributed code (before it's transpiled). A contract test against the distributed API would also be great, but it seems like this repo is performing unit tests against the distributable which I don't think is proper. The unit tests probably also shouldn't be distributed with the runtime package.

Shadouts:
1. I merged master with the typescript test changes
2. I changed fees test to a typescript module
3. I added further cleanup to the lib package by removing javascripts types modules (they were empty modules) and removed test type declarations. Types declarations will now be part of running `compile`.

src/fees.test.js
Shadouts: Changed in `f4e01f4ed2b753c0991fd6f852a001f66f74a508`

bucko13 (approved):

PR #92, opened 2023-08-23
chore(tests): add minimal typescript support for test files

bucko13:
Longer term we'll want to be properly running tests via typescript and not have to run through the full compilation. (see thread [here](https://github.com/unchained-capital/unchained-bitcoin/pull/89#discussion_r1303316602) for more context). 

In the meantime, this lets us incrementally migrate individual test files to typescript and gives a smidgeon of additional guarantees around both the typescript and the tests.


Shadouts (changes requested):
One change and I think it's good:
`compile` and `compile:watch` should both have the following args added to their babel script:
`--ignore '**/*.test.js' --ignore '**/*.test.ts'`

Shadouts (approved):

PR #94, opened 2023-08-31
fix: restore exporting of type modules for commonjs compatibility

Shadouts:
There was a regression in recent merges which causes erroring imports in JS environments. The problem is to do with how typescript is exporting types. This change restores the empty type modules to ensure commonjs compatibility.


bucko13 (approved):

bucko13:
Thanks!

PR #95, opened 2023-09-15
Typescript and v1

Shadouts:
We're upgrading everything for typescript!


src/braid.ts
```diff
@@ -294,37 +254,29 @@ export function deriveMultisigByIndex(braid, index) {
...
- * @param {Object[]} bip32Derivation this is the array of bip32info for each member of the multisig
- * @returns {module:multisig.Multisig} braid-aware MULTISIG object
  */
 function generateBraidAwareMultisigFromPublicKeys(
   braid,
   pubkeys,
   bip32Derivation
-) {
+): any {
+  let braidAwareMultisig = {};
```
bucko13, line 263: good stuff!

src/networks.ts
```diff
@@ -4,74 +4,26 @@ import { networks } from "bitcoinjs-lib";
...
-  TESTNET,
-  REGTEST,
-} as const;
+/* eslint-disable no-shadow */
+export enum Network {
+  MAINNET = "mainnet",
+  TESTNET = "testnet",
+  REGTEST = "regtest",
+  SIGNET = "signet",
+}
```
bucko13, line 14: Could we not keep the individual exports as well? It would be nice to still be able to do `import { MAINNET, TESTNET } from "unchained-bitcoin" ` So we could probably update to this: 

```typescript
export const MAINNET = "mainnet";

export enum Network {
  MAINNET = MAINNET,
  // ... etc.
}
```

src/fixtures.ts
```diff
@@ -1281,23 +1281,28 @@ const MULTISIGS_BASE = [
...
+  let braidAwareMultisig = {};
   const multisig = generateMultisigFromPublicKeys(
     test.network,
     test.type,
     2,
     ...test.publicKeys
   );
-  multisig.braidDetails = braidConfig(test.braidDetails);
-  multisig.bip32Derivation = test.bip32Derivation;
+  braidAwareMultisig = {
```
bucko13, line 1291: is this defined as a type anywhere? (sorry still reviewing but could be useful)

src/inputs.ts
```diff
@@ -46,25 +34,25 @@ export function sortInputs(inputs) {
...
   for (let inputIndex = 0; inputIndex < inputs.length; inputIndex++) {
     const input = inputs[inputIndex];
-    if (braidRequired && input.multisig && !multisigBraidDetails(input.multisig)) {
-        return `At least one input cannot be traced back to its set of extended public keys.`;
+    if (
+      braidRequired &&
+      input.multisig &&
+      !multisigBraidDetails(input.multisig)
+    ) {
+      return `At least one input cannot be traced back to its set of extended public keys.`;
```
bucko13, line 50: surprised the linter didn't get angry at this but I don't think this should have backticks if there is no interpolation
```suggestion
      return "At least one input cannot be traced back to its set of extended public keys.";
```

bucko13 (commented):
Not done yet but looking good so far. Have a question about the network module that I wanted to submit for now with a couple other comments.

src/networks.ts
Shadouts: I think keeping the individual exports will complexify the typing. If an enum value is expected, then providing a string constant will not be acceptable. There was also a comment that was removed suggesting that the constants could be removed after enums were built.

src/fixtures.ts
Shadouts: No. The "Multisig" type was only previously typed by jsdoc and even then it never had a defined shape. I think further work could be done to make it a real type, but it didn't fit in the scope of this upgrade. It seems like it would take some discovery.

src/addresses.ts
```diff
@@ -0,0 +1,44 @@
...
+const ADDRESS_BODY_PATTERN = "[A-HJ-NP-Za-km-z1-9]+$";
+const BECH32_ADDRESS_MAGIC_BYTE_REGEX = /^(tb|bc)/;
+const BECH32_ADDRESS_BODY_PATTERN = "[ac-hj-np-z02-9]+$";
+
+/**
+ * Validate a given bitcoin address.
+ *
+ * Address must be a valid address on the given bitcoin network.
+ */
+export function validateAddress(address: string, network) {
```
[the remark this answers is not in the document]
Shadouts: Yes. I think some/most were upgraded already. This might have just been overlooked.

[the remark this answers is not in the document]
Shadouts: changed in 4eb4a92681437fd9ae79b007644ce2f7e351c3b6

src/inputs.ts
Shadouts: changed in 879f0e8f5de01e469a3092e62ea6b1854732662d

src/networks.ts
Shadouts: What's the real goal here, a flexible API or a strict API? Enum ads strictness with a helpful way to reference. Union literals would be the most strict without helpful reference. If the goal is to make it the most strict without making the type confusing, I'd argue that the best would be a union literal that is not exported from the lib. Lib consumers would be required to pass a string matching one of the values specified by docs/IDE. This would probably have the least confusing typing.

bucko13: Yeah, all fair. I think was just thinking if there would be a way to minimize the diff. But this is definitely cleaner overall I think.

src/networks.test.ts
```diff
@@ -0,0 +1,53 @@
...
+} from "./networks";
+
+import { networks } from "bitcoinjs-lib";
+
+describe("networks", () => {
+  describe("networkLabel", () => {
+    it("returns a human-readable network name", () => {
+      expect(networkLabel(Network.MAINNET)).toBe("Mainnet");
+      expect(networkLabel(Network.TESTNET)).toBe("Testnet");
+      expect(networkLabel("foobar")).toBe("Testnet");
```
bucko13, line 15: lol wat

src/p2sh.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * This module provides functions and constants for the P2SH address type.
+ */
+
+/**
+ * Address type constant for "pay-to-script-hash" (P2SH) addresses.
+ */
+export const P2SH = "P2SH";
```
bucko13, line 8: I guess this is something you were trying to fix too, like Network where we could have `AddressType.P2SH`

src/transactions.test.ts
```diff
@@ -128,7 +128,7 @@ describe("transactions", () => {
             fixture.inputs,
             fixture.outputs
           );
-          expect(fixture.psbt).toEqual(psbt.toBase64());
+          expect(fixture.psbt).toEqual(psbt.data.toBase64());
```
bucko13, line 131: was this just broken? I don't see where the return of this function was changed

src/transactions.ts
```diff
@@ -0,0 +1,296 @@
...
+//  */
+// export function signedMultisigPSBT(network, inputs, outputs, transactionSignatures) {
+//   const psbt = unsignedMultisigPSBT(network, inputs, outputs);
+//  const unsignedTransaction = unsignedTransactionObjectFromPSBT(psbt); // validates inputs & outputs
+//  if (!transactionSignatures || transactionSignatures.length === 0) { throw new Error("At least one transaction signature is required."); }
+//
+//  transactionSignatures.forEach((transactionSignature, transactionSignatureIndex) => {
+//    if (transactionSignature.length < inputs.length) {
+//      throw new Error(`Insufficient input signatures for transaction signature ${transactionSignatureIndex + 1}: require ${inputs.length}, received ${transactionSignature.length}.`);
+//    }d
```
bucko13, line 264:
```suggestion
//    }
```

src/types/networks.ts
```diff
@@ -1,10 +1,4 @@
...
-const NETWORKS_ENUM = {
-  ...NETWORKS,
-  REGTEST,
-  SIGNET,
-} as const;
-
-type NETWORKS_KEYS = keyof typeof NETWORKS_ENUM;
-export type BitcoinNetwork = (typeof NETWORKS_ENUM)[NETWORKS_KEYS];
+type NETWORKS_KEYS = keyof typeof Network;
+export type BitcoinNetwork = (typeof Network)[NETWORKS_KEYS];
```
bucko13, line 4: probably don't even need this anymore tbh.

bucko13 (approved):
I think we should be good to go with this.

src/networks.test.ts
Shadouts: This existed as previous functionality. It was just moved over from another file. It's testing the friendly naming for different networks and that it defaults to "Testnet".

src/p2sh.ts
Shadouts: Yes. A change on this was attempted, but the scope is greater than Networks. While chasing it all down, I didn't complete it, and I left it out. It could still be added in if you feel it's worth it.

src/types/networks.ts
Shadouts: It was left for backwards compatibility purpose. Removing it would be breaking.

src/transactions.test.ts
Shadouts: Not sure. It's possible the fixture changed. @nk1tz ?

nk1tz: Well what's clear is that there is no `toBase64()` method on the output of `unsignedMultisigPSBT` but `.data` property is typed as `PSBT` which does have the method to do this
```
export declare class Psbt {
  ....
    toBase64(): string;
...
}
```
So perhaps there is/was a sneaky pass through previously? But typescript didn't like it.

nk1tz: `unsignedMultisigPSBT` went from:
```
export function unsignedMultisigPSBT(network, inputs, outputs, includeGlobalXpubs=false) {
  const multisigInputError = validateMultisigInputs(inputs, true);
  assert(!multisigInputError.length, multisigInputError);
  const multisigOutputError = validateOutputs(network, outputs);
  assert(!multisigOutputError.length, multisigOutputError);

  const psbt = new Psbt({ network: networkData(network) });
  // FIXME: update fixtures with unsigned tx version 02000000 and proper signatures
  psbt.setVersion(1); // Our fixtures currently sign transactions with version 0x01000000
  const globalExtendedPublicKeys = [];

  inputs.forEach((input) => {
    const formattedInput = psbtInputFormatter({...input});
    psbt.addInput(formattedInput);
    const braidDetails = input.multisig.braidDetails;
    if (braidDetails && includeGlobalXpubs) {
      const braid = Braid.fromData(JSON.parse(braidDetails));
      braid.extendedPublicKeys.forEach(extendedPublicKeyData => {
        const extendedPublicKey = new ExtendedPublicKey(extendedPublicKeyData);

        const alreadyFound = globalExtendedPublicKeys.find(
          (existingExtendedPublicKey) => existingExtendedPublicKey.toBase58() === extendedPublicKey.toBase58()
        );

        if (!alreadyFound) {
          globalExtendedPublicKeys.push(extendedPublicKey);
        }
      });
    }
  });

  if (includeGlobalXpubs && globalExtendedPublicKeys.length > 0) {
    const globalXpubs = globalExtendedPublicKeys.map(extendedPublicKey => ({
        extendedPubkey: extendedPublicKey.encode(),
        masterFingerprint: Buffer.from(extendedPublicKey.rootFingerprint, 'hex'),
        path: extendedPublicKey.path,
      })
    );
    psbt.updateGlobal({globalXpub: globalXpubs});
  }
```

to 
```
export function unsignedMultisigPSBT(
  network,
  inputs,
  outputs,
  includeGlobalXpubs = false
) {
  const multisigInputError = validateMultisigInputs(inputs, true);
  assert(!multisigInputError.length, multisigInputError);
  const multisigOutputError = validateOutputs(network, outputs);
  assert(!multisigOutputError.length, multisigOutputError);

  const psbt = new Psbt({ network: networkData(network) });
  // FIXME: update fixtures with unsigned tx version 02000000 and proper signatures
  psbt.setVersion(1); // Our fixtures currently sign transactions with version 0x01000000
  const globalExtendedPublicKeys: ExtendedPublicKey[] = [];

  inputs.forEach((input) => {
    const formattedInput = psbtInputFormatter({ ...input });
    psbt.addInput(formattedInput);
    const braidDetails = input.multisig.braidDetails;
    if (braidDetails && includeGlobalXpubs) {
      const braid = Braid.fromData(JSON.parse(braidDetails));
      braid.extendedPublicKeys.forEach((extendedPublicKeyData) => {
        const extendedPublicKey = new ExtendedPublicKey(extendedPublicKeyData);

        const alreadyFound = globalExtendedPublicKeys.find(
          (existingExtendedPublicKey: any) =>
            existingExtendedPublicKey.toBase58() ===
            extendedPublicKey.toBase58()
        );

        if (!alreadyFound) {
          globalExtendedPublicKeys.push(extendedPublicKey);
        }
      });
    }
  });

  if (includeGlobalXpubs && globalExtendedPublicKeys.length > 0) {
    const globalXpubs = globalExtendedPublicKeys.map((extendedPublicKey) => ({
      extendedPubkey: extendedPublicKey.encode(),
      masterFingerprint: extendedPublicKey.rootFingerprint
        ? Buffer.from(extendedPublicKey.rootFingerprint, "hex")
        : Buffer.alloc(0),
      path: extendedPublicKey.path || "",
    }));
    psbt.updateGlobal({ globalXpub: globalXpubs });
  }

  const psbtOutputs = outputs.map((output) =>
    psbtOutputFormatter({ ...output })
  );
  psbt.addOutputs(psbtOutputs);
  const txn = psbt.data.globalMap.unsignedTx.toBuffer().toString("hex");

  return { ...psbt, txn };
}
```

So I don't think that explains it either.

src/types/networks.ts
bucko13: I guess I just figured since we're assumign this whole thing was going to be breaking we could get away with it, but not a huge deal either way.

src/networks.test.ts
bucko13: Yeah I know. I'm just kind of laughing that this is/was supported to begin with.

bucko13 (approved):

PR #100, opened 2023-10-27
chore: swap travis for github actions

waldenraines:


bucko13 (approved):
Seems reasonable enough!

PR #106, opened 2023-12-14
Fix validateOutputAmount API for BigNumber

Shadouts:
`outputs/validateOutputAmount` is requiring a BigNumber type on its args. This leaks lib dependencies onto the caller. Previously a version of unchained-bitcoin had removed the BigNumber dependency from its API. This function must have been missed.

Unfortunately this change would require a major version increment since an arg type is updating and breaking backward compatibility.

### DIFF
--- a/src/outputs.ts
+++ b/src/outputs.ts
@@ -61,7 +61,7 @@ export function validateOutput(network, output, inputsTotalSats?) {
 /**
  * Lowest acceptable output amount in Satoshis.
  */
-const DUST_LIMIT_SATS = new BigNumber(546);
+const DUST_LIMIT_SATS = "546";
 
 /**
  * Validate the given output amount (in Satoshis).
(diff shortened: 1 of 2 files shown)


src/outputs.ts
```diff
@@ -61,7 +61,7 @@ export function validateOutput(network, output, inputsTotalSats?) {
 /**
  * Lowest acceptable output amount in Satoshis.
  */
-const DUST_LIMIT_SATS = new BigNumber(546);
+const DUST_LIMIT_SATS = new BigNumber(546).toString();
```
bucko13, line 64: Couldn’t this just be 
```suggestion
const DUST_LIMIT_SATS = “546”
```

Shadouts: Yes. I don't think the difference matters that much. I think there is a difference in that the first is like a casting of an int and the second is purely a string which may not have started as an int.

Shadouts:
I've changed this to offer backward compatibility on the function. Any future major version should remove the BigNumber type from the arg.

bucko13 (approved):

PR #107, opened 2023-12-14
Adjust node versions to match minimum supported

Shadouts:
This change adjusts github workflows to match the minimum supported node version. `package.json` indicates that the minimum supported version of node is >= 16, but ci was building for 14 (incompatible) and release-please was building using 18. 

If this change is unacceptable, I would recommend alternatively `package.json` engines entry be updated to >=18 and ci strategy node versions range `['18.x', '20.x']`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['14.x', '16.x', '18.x']
+        node-version: ['16.x', '18.x', '20.x']
     
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


bucko13:
> If this change is unacceptable, I would recommend alternatively package.json engines entry be updated to >=18 and ci strategy node versions range ['18.x', '20.x']

This feels preferable to me honestly especially given that [Node 16 is past EoL](https://nodejs.org/en/blog/announcements/nodejs16-eol). Was there a reason to try this first or was it just to err on the side of being conservative?

Shadouts:
> > If this change is unacceptable, I would recommend alternatively package.json engines entry be updated to >=18 and ci strategy node versions range ['18.x', '20.x']
> 
> This feels preferable to me honestly especially given that [Node 16 is past EoL](https://nodejs.org/en/blog/announcements/nodejs16-eol). Was there a reason to try this first or was it just to err on the side of being conservative?

Not upgrading maintains backward compatibility. If 16 gets dropped on this lib, then npm will yell when someone tries to install to their project using 16.

bucko13 (approved):

PR #109, opened 2024-01-03
fix(fixtures): fix hmac fixtures

bucko13:
Fixtures didn't match wallet configs. Some earlier change must have broken this and the fixtures weren't caught as well. With this change, all relevant tests in the caravan test suite pass (nest segwit signing still doesn't work but that's not related, it was failing w/ a different error previously though) 

Screenshot 2024-01-03 at 2 28 10 PM

### DIFF
--- a/src/fixtures.ts
+++ b/src/fixtures.ts
@@ -558,7 +558,7 @@ const MULTISIGS_BASE = [
     type: P2SH,
     bip32Path: "m/45'/1'/100'/0/0",
     policyHmac:
-      "6fb498be86e7accd2b264c6d129e2b1bf5edad95dbc194db962ca8958c4af0c9",
+      "fb633e6a50ff05a1a090e07f2586d11f04b4ee8052cd0ce9ed816f8c4446cdc3",
     publicKey:
       "02a8513d9931896d5d3afc8063148db75d8851fd1fc41b1098ba2a6a766db563d4",
     publicKeys: [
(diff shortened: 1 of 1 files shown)


dylan-thompson:
I'm not sure if this is related but for all the fixtures you updated key 0 (as displayed in caravan) is shown as key 1 when registering the wallet on my ledger and vice-versa. 

Also, the tests that are failing for you work for me. Is your bitcoin testnet app up to date?

dylan-thompson:
The only ledger test failing for me is `Export mainnet xpub at m/45'/0'/4'/99'/2147483647/3/1` due to the derivation path being too long. Should we keep the test or remove it?

bucko13:
> I'm not sure if this is related but for all the fixtures you updated key 0 (as displayed in caravan) is shown as key 1 when registering the wallet on my ledger and vice-versa.

Ah, this might've been what changed the hmac. We weren't sorting the keys before and now we do to make the registrations less brittle. Might be best to change that in caravan (sort them before showing them in the table)

> Also, the tests that are failing for you work for me. Is your bitcoin testnet app up to date?
Very possible it's not! Good to hear that his got fixed though. I'll try again.

bucko13:
> The only ledger test failing for me is Export mainnet xpub at m/45'/0'/4'/99'/2147483647/3/1 due to the derivation path being too long. Should we keep the test or remove it?

Hard to say. Possible they'll start allowing this eventually. Also not all devices will fail for this I don't think... Might be something to consider for the test suite, and maybe when/if we rebuild it to be more customizable and robust for these types of edge cases. Like, I think it's good to have but the test suite should support flagging tests like this.

dylan-thompson (approved):

PR #110, opened 2024-01-04
Fee error types

Shadouts:
In order to provide more flexible uses for fee validation, this change introduces fee error types.

Previously, fee validation only returned strings. With validation functions which return an error type, applications using this library will be able to provide their own user feedback based on error type.

The new type `FeeValidationError` is a typescript enum collecting fee validation error categories currently being returned by `validateFee` and `validateFeeRate` functions.

To preserve backward compatibility, `validateFee` and `validateFeeRate` functions are now merely wrappers for more granular functions which return pieces of an error. These new functions are: `checkFeeError`, `checkFeeRateError`, and `getFeeErrorMessage`.

I'm not entirely happy with the naming. I think the ideal names for the more "raw" validation feedback that `checkFeeError` and `checkFeeRateError` perform would have used "validate", but validate was already used for the existing functions which return the error message. I'm open to suggestions.

### DIFF
--- a/src/fees.ts
+++ b/src/fees.ts
@@ -141,7 +141,7 @@ export function checkFeeRateError(feeRateSatsPerVbyte) {
  * Validate the given transaction fee rate (in Satoshis/vbyte). Returns an
  * error message if invalid. Returns empty string if valid.
  *
- * - Must be a parseable as a number.
+ * - Must be parseable as a number.
  *
  * - Cannot be negative (zero is OK).
  *
--- a/src/fees.ts
+++ b/src/fees.ts
@@ -5,6 +5,8 @@
 
 import BigNumber from "bignumber.js";
 
+import { FeeValidationError } from "./types";
+
 import { P2SH, estimateMultisigP2SHTransactionVSize } from "./p2sh";
 import {
   P2SH_P2WSH,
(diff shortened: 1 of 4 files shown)


src/fees.ts
```diff
@@ -28,76 +30,141 @@ const MAX_FEE_RATE_SATS_PER_VBYTE = new BigNumber(1000); // 1000 Sats/vbyte
...
-    return "Fee rate cannot be negative.";
-  }
-  if (fr.isGreaterThan(MAX_FEE_RATE_SATS_PER_VBYTE)) {
-    return "Fee rate is too high.";
+export function getFeeErrorMessage(error: FeeValidationError | null) {
+  let errorMessage = "";
+
+  switch (error) {
+    case FeeValidationError.FEE_CANNOT_BE_NEGATIVE:
+      errorMessage = "Fee cannot be negative.";
```
bucko13, line 40: Can't these just return the string at the end of each case?

src/fees.ts
```diff
@@ -28,76 +30,141 @@ const MAX_FEE_RATE_SATS_PER_VBYTE = new BigNumber(1000); // 1000 Sats/vbyte
...
+  }
+
+  return null;
+}
+
+/**
+ * Validate the given transaction fee rate (in Satoshis/vbyte). Returns an
+ * error message if invalid. Returns empty string if valid.
+ *
+ * - Must be a parseable as a number.
```
bucko13, line 144:
```suggestion
 * - Must be parseable as a number.
```

bucko13 (changes requested):
Couple small nits. Overall I like the change. I see what you mean about the names, but I'm fine with what you've got for now until we're ready to make breaking changes.

src/fees.ts
Shadouts: They could be, but I don't like using returns in switch statements. There is already a conventional command to interrupt switch statement fall-through. Introducing a different one isn't needed and may result in a bigger change in the future if modifying the switch statement or function. If this single exit-point pattern isn't acceptable, I'd probably change the function to use if statements, but I think the switch is more performant when checking this type.

bucko13 (approved):

src/fees.ts
tylerchambers: could I convince you to use an object mapping here then?
```ts
const feeErrorMessages = {
  [FeeValidationError.FEE_CANNOT_BE_NEGATIVE]: "Fee cannot be negative.",
  [FeeValidationError.FEE_RATE_CANNOT_BE_NEGATIVE]: "Fee rate cannot be negative.",
  // ... other mappings
};

export function getFeeErrorMessage(error: FeeValidationError | null): string {
  return error !== null ? feeErrorMessages[error] ?? "" : "";
}
```

tylerchambers: oh nevermind lol

bucko13: whoops. sorry.

bucko13: Nothing released yet so we can revert

bucko13: The mapping is nice.

tylerchambers: Six of one, a half dozen of the other. It's all the same to me, really.

